PR #252, opened 2012-12-02
Split SimplePie (the class) into SimplePie and SimplePie_Feed

rmccue:
SimplePie (the class) should be the main point of entry for users, but not where feed-level data is stored. This splits it out into a separate class instead.

This is not yet ready. We should probably also look at `SimplePie_Source`.

### DIFF
--- a/library/SimplePie.php
+++ b/library/SimplePie.php
@@ -683,6 +683,62 @@ class SimplePie
 	}
 
 	/**
+	 * Set an instance of {@see SimplePie_File} to use as a feed
+	 *
+	 * @param SimplePie_File &$file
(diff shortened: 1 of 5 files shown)


mblaney:
This is a great idea, hopefully someone is willing to take it on some day. Thanks for all your efforts with SimplePie @rmccue, I'll close this for now but if anyone else is keen for a major refactor I'm happy to review :-)

skyzyx:
Honestly, I've been poking at this in the background for the past couple of months. I'll push my development branch back to here once I make a little more progress.

mblaney:
@skyzyx awesome look forward to seeing that :-)

Issue #260, opened 2013-01-15
Growing memory consumption on get_links (including fix!)


h2ooooooo:
Problem question and fix on this SO post:

http://stackoverflow.com/questions/14337179/why-am-i-getting-memory-leaks-in-simplepie-when-using-item-get-permalink/

talgalili:
I believe this issue is now resolved, and should be closed. See:

https://github.com/simplepie/simplepie/commits/master/library/SimplePie/IRI.php

PR #287, opened 2013-04-12
Fix for Issue #260 — Memory Consumption

kkirsche:
Issue: https://github.com/simplepie/simplepie/issues/260

Fix: http://stackoverflow.com/questions/14337179/why-am-i-getting-memory-leaks-in-simplepie-when-using-item-get-permalink/


kkirsche:
Any luck with this guys?

PR #290, opened 2013-04-24
Russian characters fix

jagermesh:
To have

image

Instead of 

image


mblaney:
Hi @jagermesh do you have an example feed that shows what's going on here?

Issue #343, opened 2014-01-30
Generate id as hash over link, title and content if GUID not present


BernhardPosselt:
Linking this thread: https://github.com/owncloud/news/issues/217

PR #344, opened 2014-01-31
use hash over title, content and link if guid not given

BernhardPosselt:
fix for #343

Did not test this, just editted like i thought that it should work, please review

### DIFF
--- a/library/SimplePie/Item.php
+++ b/library/SimplePie/Item.php
@@ -238,19 +238,8 @@ class SimplePie_Item
 			{
 				return $this->sanitize($this->data['attribs'][SIMPLEPIE_NAMESPACE_RDF]['about'], SIMPLEPIE_CONSTRUCT_TEXT);
 			}
-			else
-			{
-				return md5(serialize($this->data));
-			}
 		}
-		if ($this->get_permalink() !== null || $this->get_title() !== null)
-		{
-			return md5($this->get_permalink() . $this->get_title());
-		}
-		else
-		{
-			return md5(serialize($this->data));
-		}
+		return md5(serialize($this->data));
 	}
 
 	/**
--- a/library/SimplePie/Item.php
+++ b/library/SimplePie/Item.php
@@ -238,13 +238,9 @@ class SimplePie_Item
 			{
 				return $this->sanitize($this->data['attribs'][SIMPLEPIE_NAMESPACE_RDF]['about'], SIMPLEPIE_CONSTRUCT_TEXT);
 			}
-			elseif (($return = $this->get_permalink()) !== null)
+			else
 			{
-				return $return;
-			}
-			elseif (($return = $this->get_title()) !== null)
-			{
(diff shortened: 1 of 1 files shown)


mblaney:
@dhilowitz sorry to hear that. I've just started using get_id myself, and see that it is fundamentally broken if it reaches serialize($this->data). This is because the 'build' timestamp is included in the data, so if SimplePie changes all your hashed id's will change...

I propose we change the line to:
md5($this->get_permalink().$this->get_title().$this->get_content())

Which was one of the previous code paths... note that the current line has always been there, so it was always possible to generate faulty hashes when the build changed.

Not sure how to mitigate duplicates in doing this, but open to suggestions. It shouldn't actually be a problem, I also store items in a database and just remove old duplicates based on the permalink.

JanPetterMG:
I've been struggling with this myself the last couple of years. This is what works best for me and my database of about 732.000 rss/atom feeds from all over the world:

Whatever comes first.
1. use GUID (if it is an md5 hash)
2. use an md5 of the GUID (when it's not a hash)
3. use an md5 of the link (if a link exists)
4. use an md5 of the title and description (when both exists)
5. use an md5 of the title (as last option)

I don't say the solution used by simplepie is wrong, just that anyone parsing feeds seriously and for commercial reasons, should take this part very seriously.

My experience is that item titles change rapidly, so does the description. The result is large amounts of duplicate items in the database, screwing up the list of the newest items.

I would recommend creating a function crunching all the data, witch returns an md5. It's actually pretty simple using if else statements, just a few extra lines of code needed :-)

What do you folks think?

mblaney:
Thanks for the feedback @JanPetterMG. Did you see the original discussion in #343? Using `link` when `guid` was not set was the reason for this change in the first place, because that caused it's own problems.

I would prefer that SimplePie only returned explicit id's found in the feed and let custom code such as your own resolve what to do, but maybe supporting existing implementations is important in this case.

mblaney:
I'm going to make the change to get_id() that I proposed above, just a heads up that it will cause more duplicates if you haven't coded against them. I also think the current version is broken enough that it warrants a new 1.4.1 release, but will of course include other recent changes.

Will look at doing this in about a week, so if you have any pending changes you want included now would be a good time to add them. My personal interest at the moment is in microformats feeds, so will be adding a few improvements in that area too.

thinkingmedia:
@mblaney while you are at that part of the code. Can you add an extra parameter for a callable. If a callable is passed that is used to generate the hash. That way we can replace the `md5` call with something else.

mblaney:
@thinkingmedia that's a good idea. Let me know if you want to be able to distinguish between permalink, title and content too (ie I could add newlines between them). Otherwise I'll change it to this:
```
if ($fn === '' || !is_callable($fn)) $fn = 'md5';
return call_user_func($fn, $this->get_permalink().$this->get_title().$this->get_content());
```

PR #351, opened 2014-02-17
[FreshRSS] New function add_attributes()

Alkarex:
Just like there is a strip_attributes() function, here is a new
add_attributes() which works on the same model, except it is to add
attributes instead of removing them.

Just like strip_attributes(), add_attributes() comes with some default
settings, namely to add a few HTML5 attributes:
```html
<audio preload="none" />
<video preload="none" />
<iframe sandbox="allow-scripts allow-same-origin" />
```

Watch out that I have changed the signature of the protected functions
strip_tag() and strip_attr() for performance reasons (to avoid a loop of
new DOMXPath()). If this is a problem, I could propose a patch with the
old signatures.

This is a first attempt to report upstream some improvements made for
the FreshRSS news reader http://freshrss.org. If there is interest,
there will be more.


Alkarex:
Sorry, I deleted this pull request by mistake. I have just made a new, identical, pull request https://github.com/simplepie/simplepie/pull/394

PR #359, opened 2014-04-07
Implement proxy variables for curl_setopt

nerzhul:
Updated File.php since discussion on issue https://github.com/simplepie/simplepie/pull/357#issuecomment-39635250


Alkarex:
See also https://github.com/simplepie/simplepie/pull/407

mblaney:
fixed by #407

PR #360, opened 2014-04-07
Implement proxy variables for curl_setopt (part 2)

nerzhul:
Linked with https://github.com/simplepie/simplepie/pull/359


Alkarex:
See also See also https://github.com/simplepie/simplepie/pull/407

mblaney:
fixed by #407

PR #364, opened 2014-04-29
Compile new simplepie

mkly:


mblaney:
I'm going to close this then, thanks for 6f76136 @mkly

PR #371, opened 2014-07-28
-fix bug in IRI get_iri

simion314:
for the url 
http://news.softpedia.com/newsRSS/Linux-7.xml
 it returned the wrong url
http://news.softpedia.com/newsRSS/Linux-7.xml?#
 , it added the extra "?#" at the end


melon-babak:
Any updates on this? I may be able to contribute a patch but I'm not clear on why the ```(string)``` casts are there in the first place. It would be great if someone more familiar with the codebase could provide some direction about a feasible fix.

Issue #334, opened 2013-12-09
Decoding HTML entities


d-ogarkov:
Assuming that SimplePie serves international UTF-8 characters as HTML entities by nature, and there is no way to change this behaviour at this moment (I'm using 1.3.1), we're forced to use html_entity_decode on every piece of data we receive from SimplePie.

If we skip this step, we'll have every single letter encoded with 7 (seven) bytes where it's possible to use just two - this is very ineffective.

This is a feature request for full UTF-8 support in future versions.

lexvjatkin:
I made a cleaner solution for this issue (test included). If you need something else to add just let me know.

PR #378, opened 2014-09-15
Fix issue 334

Hubbitus:
Fix of https://github.com/simplepie/simplepie/issues/334#issuecomment-55569775


Alkarex:
https://github.com/simplepie/simplepie/pull/470

Issue #379, opened 2014-10-14
Missing CEST timezone


barbuslex:
Hi,

In SimplePie_Parse_Date class it missing CEST time zone.

I fix it in pull request.

Thanks to solve it

ifsnop:
Merged in c78d401, also Closes #380

PR #380, opened 2014-10-14
add CEST timezone

barbuslex:
add missing CEST timezone

according to the issue #379


library/SimplePie/Parse/Date.php
```diff
@@ -332,6 +332,7 @@ class SimplePie_Parse_Date
 		'CDT' => -18000,
 		'CEDT' => 7200,
 		'CET' => 3600,
```
Alkarex: `CEST` is before `CET` in the alphabetic order.

PR #387, opened 2015-01-12
Fixed superfluous entities in Sanitizer + unit test

lexvjatkin:
Better solution for simplepie/simplepie#334 and simplepie/simplepie#290 
(html export is corruped (bad encoding) when document head is removed/replaced)

 * HTML export in DOMDocument is fine, we need just extract content after it's done with right headers
 * added unit test for Sanizite class


lexvjatkin:
Come on, PHP 5.2 support? It's dead for 4 years now http://php.net/eol.php

PR #388, opened 2015-01-26
Sanitize: superfluous entities

lexaurin:
Please see https://github.com/simplepie/simplepie/issues/334 and https://github.com/simplepie/simplepie/pull/290 
This is complete solution / pull-request for this bug. 

Included:
  * unit test
  * bump off PHP 5.2
  * fix for PHP 5.3 by jagermesh@d91ae5e718a446652efd08cb42a622c93226d215


mblaney:
Hi @lexvjatkin thanks for this, but I think it's already been fixed. Have you tried https://github.com/simplepie/simplepie/pull/378?

I ran your new unit test and it passes with this, which means support for php 5.2 doesn't need to be dropped.

Alkarex:
@mblaney Now that SimplePie does not support PHP 5.2 anymore, I think this approach looks cleaner and more efficient than https://github.com/simplepie/simplepie/pull/378/files
It requires PHP 5.3.6+ though. SimplePie might still work with PHP 5.3.0, but unsure, and Travis is testing against PHP 5.3.29.

mblaney:
I've had problems with the existing approach, and had to make some changes to it, so if you think the other method is better I'm happy to try it too. Do you want to make a clean patch that can be merged?

Alkarex:
@mblaney Here you are https://github.com/simplepie/simplepie/pull/470

PR #398, opened 2015-04-05
[FreshRSS] HTTP 301 Moved Permanently

Alkarex:
Add support for HTTP 301 Moved Permanently.

This patch implements one of SimplePies TODO's: "If we have a perm
redirect we should return the new URL".
https://github.com/simplepie/simplepie/blob/43b1018e8df4a16d6573968cdbcbcffdd6bab3e3/library/SimplePie.php#L1747

https://github.com/FreshRSS/FreshRSS/commit/63c9c47a761b9925e6811ae630c3b77ef05b9fc8

### DIFF
--- a/library/SimplePie.php
+++ b/library/SimplePie.php
@@ -446,6 +446,13 @@ class SimplePie
 	public $feed_url;
 
 	/**
+	 * @var string Original feed URL, or new feed URL iff HTTP 301 Moved Permanently
+	 * @see SimplePie::subscribe_url()
+	 * @access private
+	 */
+	public $permanent_url = null;
+
+	/**
 	 * @var object Instance of SimplePie_File to use as a feed
 	 * @see SimplePie::set_file()
 	 * @access private
@@ -728,6 +735,7 @@ class SimplePie
 		else
 		{
 			$this->feed_url = $this->registry->call('Misc', 'fix_protocol', array($url, 1));
+			$this->permanent_url = $this->feed_url;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


mblaney:
Hi @Alkarex I was looking at File.php today and noticed your changes may not do exactly what you're after. Unless you've got an older version of curl, it will do the redirects itself and SimplePie only parses the last headers. To get the previous location you would need to get the last set of headers too and save the Location. Just wanted to let you know.

Alkarex:
Hello @mblaney and thanks for the feedback. I still think it works as expected. I do keep the previous location before calling cURL: https://github.com/Alkarex/simplepie/blob/HTTP301/library/SimplePie/File.php#L78 Please let me know if you still think there is a problem. It seems to work fine in FreshRSS at least.

mblaney:
I must be missing something then. If that's the only place you set $this->permanent_url, then how does it ever differ from $this->url?

Alkarex:
@mblaney See also https://github.com/Alkarex/simplepie/blob/HTTP301/library/SimplePie/File.php#L147

mblaney:
yeah that's why I sent the first message, that line never gets called. curl will do the redirects internally, and then SimplePie pops the last set of headers so you will only see a status code of 200. I was trying to do something similar to your permanent url feature when I noticed this.

Alkarex:
@mblaney You mean that the whole section https://github.com/Alkarex/simplepie/blob/HTTP301/library/SimplePie/File.php#L141 is never called because the status code is never 30x?

mblaney:
exactly, any 30x's will be further down the headers array and SimplePie only uses the last one, which will most likely have a status code of 200.

PR #401, opened 2015-04-06
[FreshRSS] MD5-based caching

Alkarex:
_This patch may not be suited to be merged as-is in SimplePie, but it is used in FreshRSS since February 2014, and may contribute to a discussion/process of improving HTTP caching in SimplePie. It is a part of the process of proposing/reporting upstream the changes made to SimplePie in FreshRSS_

Introduces `data['mtime']` and `data['md5']`.

For feeds that do not support HTTP conditional requests, makes a MD5 hash to detect changes.
To do that, some cleaning is needed, e.g. removing comments such as `<!-- Generated on XX -->`, or `<lastBuildDate>` which may be changing even when no actual content has changed.
https://github.com/FreshRSS/FreshRSS/commit/9aab83af115de3b210ea41caae49b70a0f82492a

SimplePie::init() now returns a positive integer with modification time if using cache, boolean true if otherwise successful, false otherwise.

Take also care of the caching of feeds with errors. Before, the cache system was not used for feeds with errors; there was no protection against repetitive refresh.
https://github.com/FreshRSS/FreshRSS/commit/00127f07c5fc784130d658e3f26519b0279fc6b8

### DIFF
--- a/library/SimplePie.php
+++ b/library/SimplePie.php
@@ -1205,6 +1205,11 @@ class SimplePie
 		$this->enable_exceptions = $enable;
 	}
 
+	function cleanMd5($rss)
+	{
+		return md5(preg_replace(array('#<(lastBuildDate|pubDate|updated|feedDate|dc:date|slash:comments)>[^<]+</\\1>#', '#<!--.+?-->#s'), '', $rss));
+	}
+
 	/**
 	 * Initialize the feed object
 	 *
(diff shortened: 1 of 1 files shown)


mblaney:
Hi @Alkarex, I don't like the idea of changing the init() signature, you could expose mtime with a member variable if you want access to it.

PR #403, opened 2015-04-06
[FreshRSS] simplify mtime and touch

Alkarex:
filemtime already returns false if the file does not exist.
touch should work also when file does not exist yet.
https://github.com/FreshRSS/FreshRSS/commit/00127f07c5fc784130d658e3f26519b0279fc6b8

### DIFF
--- a/library/SimplePie/Cache/File.php
+++ b/library/SimplePie/Cache/File.php
@@ -136,11 +136,7 @@ class SimplePie_Cache_File implements SimplePie_Cache_Base
 	 */
 	public function mtime()
 	{
-		if (file_exists($this->name))
-		{
-			return filemtime($this->name);
-		}
-		return false;
+		return @filemtime($this->name);
 	}
 
 	/**
@@ -150,11 +146,7 @@ class SimplePie_Cache_File implements SimplePie_Cache_Base
 	 */
 	public function touch()
 	{
-		if (file_exists($this->name))
-		{
-			return touch($this->name);
-		}
-		return false;
+		return @touch($this->name);
 	}
 
 	/**


mblaney:
hi @Alkarex, nice work on all your updates! I've updated my own repository to include your changes that are passing tests, any thoughts on the ones that aren't?

On this particular PR, looks like you don't need to suppress warnings on touch()

thanks,
Malcolm

Alkarex:
Hello @mblaney and thanks. Please let me know if you have any feedback on the patches you have merged. I would also be pleased to hear more about your repository, for instance how much it is used (any running solution based on it?). Maybe you could ask to get write access to the main SimplePie repository? Looking at the queue of issues and pull-requests, some help appears to be needed :-)

Regarding https://github.com/simplepie/simplepie/pull/397 , I do not agree with the test that "//example.org/img" should be transformed into "http://example.org/img" because when the result is served by HTTPS, it will generate some mixed-content problems. It should really stay with "//example.org/img". We can discuss further in https://github.com/simplepie/simplepie/pull/397 if needed, to avoid polluting the thread to much here.

Regarding https://github.com/simplepie/simplepie/pull/400 , I do not think the tests integrate the encoding practices of e.g. Facebook and others. On the other hand, I have not yet found drawbacks of this patch (e.g. examples of feeds that worked before but not after this patch).

Regarding the warning on touch(), it may occur if the file cannot be written, but I thought it was enough to return false.

mblaney:
hi @Alkarex I just add them here https://github.com/mblaney/simplepie it contains most of the pull requests that have been submitted since the official repository stopped being maintained. I'm just using SimplePie for a personal project (unicyclic.com) so there are no other users at the moment, if there are any maintainers reading this though I would be happy to have write access :-)

If you don't agree with the tests feel free to change them, that's what I've done in the past. It also helps to clarify what an accurate test should look like given your changes.

No worries about touch(), I just didn't see any warnings listed here php.net/touch

PR #404, opened 2015-04-16
Redis cache

andre-sa:
Added Redis cache support.


mblaney:
thanks @andre-sa looks good.

AnTheMaker:
There seems to be a quite significant bug... See #616

PR #407, opened 2015-05-25
Added set_curl_options method to allow custom options

arraintxo:
We were having troubles with self-signed https hosts and have seen more people with the same issue, so we modified SimplePie to simply allow custom curl_options to be passed as configuration params.

Usage:
```php
            $simplePie = new SimplePie();
            $simplePie->set_curl_options(
                array(
                    CURLOPT_SSL_VERIFYHOST => false,
                    CURLOPT_SSL_VERIFYPEER => false
                )
            );
```

### DIFF
--- a/library/SimplePie/File.php
+++ b/library/SimplePie/File.php
@@ -65,7 +65,7 @@ class SimplePie_File
 	var $error;
 	var $method = SIMPLEPIE_FILE_SOURCE_NONE;
 
-	public function __construct($url, $timeout = 10, $redirects = 5, $headers = null, $useragent = null, $force_fsockopen = false)
+	public function __construct($url, $timeout = 10, $redirects = 5, $headers = null, $useragent = null, $force_fsockopen = false, $curl_options = array())
 	{
 		if (class_exists('idna_convert'))
 		{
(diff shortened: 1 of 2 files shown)


mblaney:
thanks @arraintxo!

library/SimplePie/File.php
```diff
@@ -112,6 +112,9 @@ public function __construct($url, $timeout = 10, $redirects = 5, $headers = null
 					curl_setopt($fp, CURLOPT_FOLLOWLOCATION, 1);
 					curl_setopt($fp, CURLOPT_MAXREDIRS, $redirects);
 				}
+                foreach ($curl_options as $curl_param => $curl_value) {
+                    curl_setopt($fp, $curl_param, $curl_value);
```
Alkarex: See http://php.net/manual/function.curl-setopt-array

Alkarex:
Hello @arraintxo,
There are a few other calls to the `File` constructor, which may also need to be updated with your new `$this->curl_options` parameter:

* [Locator.php#L251](https://github.com/BarnetikKoop/simplepie/blob/51f9f08c402d325e92a72f7568dd261c4cc329bf/library/SimplePie/Locator.php#L251)
* [Locator.php#L330](https://github.com/BarnetikKoop/simplepie/blob/51f9f08c402d325e92a72f7568dd261c4cc329bf/library/SimplePie/Locator.php#L330)
* [Locator.php#L358](https://github.com/BarnetikKoop/simplepie/blob/51f9f08c402d325e92a72f7568dd261c4cc329bf/library/SimplePie/Locator.php#L358)
* [Sanitize.php#L317](https://github.com/BarnetikKoop/simplepie/blob/51f9f08c402d325e92a72f7568dd261c4cc329bf/library/SimplePie/Sanitize.php#L317)

What do you think?

samwilson:
+1 this would be a useful feature.

PR #408, opened 2015-05-25
The "version" should never be declaired like that

GrahamCampbell:


GaryJones:
A little more context would be appreciated here I think.

GrahamCampbell:
Packagist uses git tags to work out the version. You should never specify it like that in the composer.json as it just conflicts with git creating a mess where every new commit redefines the what the release is so every commit you push is now that release.

GrahamCampbell:
This leads to a massive mess with composer's caching because it's now undefined what users actually end up with because composer was designed under the assumption releases aren't retagged.

GrahamCampbell:
Ping.

rmccue:
Thanks! :)

PR #411, opened 2015-06-28
Adds missing break statements for switch cases

pdewouters:


mblaney:
Hi @pdewouters those breaks were left out intentionally, you want the switch cases to fall through depending on how many arguments are provided.

PR #412, opened 2015-07-01
Improved error message

SteffenPegenau:
If there is an encoding error, it would be nice to know on which feed


samwilson:
+1 for this! It's super-annoying when parsing a hundred feeds to have some error somwhere and not know which feed to investigate. :)

danpoltawski:
+1 Ping.

mblaney:
fixed by 061d630

samwilson:
Thanks @mblaney :)

PR #415, opened 2015-09-04
Add Ability to Bind to a specific IP address to SimplePie

TimIgoe:


mblaney:
Hi @TimIgoe this functionality has already been added, see https://github.com/simplepie/simplepie/pull/407. Would be nice if there was a maintainer interested in dealing with PR's so effort like this didn't need to be duplicated.... thanks though, I wonder what you think of @arraintxo more generic version?

PR #419, opened 2015-10-08
Update build/compile.php to fix error

sutara79:
On Windows, the compiling fails.  
Because the argument of `file_get_contents()` is directory.  
If you confirm that the argument is NOT directory, the compiling succeeds.

```cmd
C:\simplepie>php build\compile.php

Warning: file_get_contents(C:\simplepie/library/SimplePie\.): failed to open stream: Permission denied in C:\simplepie\build\compile.php on line 54
```


mblaney:
hi @sutara79 could you check if this has been fixed by 6f76136? thanks.

Issue #423, opened 2016-01-06
pubDate vs pubdate


kaffeeguru:
I just recently stumbled over this issue of a feed providing publishing date with all lower case letters.

function SimplePie_Item::get_date only checks for 'pubDate'.

Suggest to add a check for lower case 'pubdate' as well to be more compatible to more feeds.

Example feed item I'm talking about:

<item>
	<title>
		<![CDATA[test title]]>
	</title>
	<link>http://example.com</link>
	<AuthorName/>
	<description>
		<![CDATA[content removed]]>
	</description>
	<content>
		<![CDATA[text removed]]>
	</content>
	<pubdate>2016-01-06 22:25:37</pubdate>
	<category>
		<![CDATA[Society]]>
	</category>
	<sourcename>
		<![CDATA[blabla]]>
	</sourcename>
</item>

mblaney:
the RSS spec uses camel case, so if you've only seen it in one feed I don't think it's worth adding.

PR #424, opened 2016-01-07
recognize lower case pubdate additional to pubDate

kaffeeguru:
Improvement for issue #423 please review if it makes sense


mblaney:
I don't think it's worth making a special case for one typo, unless lots of feeds do this? I haven't seen any but reopen if you really think it's an issue.

PR #434, opened 2016-03-26
Use the cache_duration variable for all types of cache

kenijo:
I made SimplePie::$cache_duration a static variable.
Then I modified the
different cache classes to access it and use it as the custom value.

### DIFF
--- a/library/SimplePie.php
+++ b/library/SimplePie.php
@@ -516,7 +516,7 @@ class SimplePie
 	 * @see SimplePie::set_cache_duration()
 	 * @access private
 	 */
-	public $cache_duration = 3600;
+	public static $cache_duration = 3600;
 
 	/**
 	 * @var int Auto-discovery cache duration (in seconds)
@@ -874,7 +874,7 @@ class SimplePie
 	 */
 	public function set_cache_duration($seconds = 3600)
 	{
-		$this->cache_duration = (int) $seconds;
+		self::$cache_duration = (int) $seconds;
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


kenijo:
Actually I'm not sure this is the right way to do it.
I get confused at how the cache works in SimplePie. It looks like $cache_duration should be the variable to be used across all cache systems but it might not work that way.

kenijo:
So I did more research and I understood that the $cache_duration is meant for HTTPCG. This is how often SimplePie will the feed through HTTPCG protocol.
This is different from the actual cache duration used for storing the data in MySQL, Memcache, etc...
Although it looks like it is tightly tied to the filecache system.

I my case of using MySQL as the cache system there are 3 options:
 - Create a new variable for the cache length independently from HTTPCG (probably ideal but requires more extensive change of the signature call to the cache function)
 - Manually modify the value of the cache length in the cache file itself (not the right way to do it)
 - Add a function outside of SimplePie in my code to manage the cache length independently from SimplePie.

So on those comment, I'll close the pull request.

ifsnop:
If you want to redefine cache duration, to be able to purge elements from cache, please take a look at the MySQL.php file.
As an optional parameter to the constructor (when you create the sample pie object), you can define this value (in seconds):

'cache_purge_time' => 2592000

Hope it helps.

kenijo:
Yes thank you. I finally figured it out.
I was hoping for the $cache_duration variable to be the one handling the purge limit but it is not the case. It would be nice to be able to change it through a variable.

In the application I am building, the MySQL cache grows beyond 100MB very quickly..
I will follow your advise through and limit it manually for now.

PR #440, opened 2016-04-23
phpDoc fixes

JanPetterMG:
For every ``get_items`` array item method, an syntax error is shown in modern php editors, ``Method not found``, Syntax and code is valid, but the editor complains due to this incorrect phpDoc return type.


JanPetterMG:
Added the same fix for ``get_enclosures``.
I don't know how many places such a fix is needed, but I'll keep adding them as I discover them...

JanPetterMG:
Documentation:
https://www.phpdoc.org/docs/latest/guides/types.html#arrays

Issue #441, opened 2016-04-25
Undefined offset: 2 in Sanitize.php


JanPetterMG:
``Notice: Undefined offset: 2 in /library/SimplePie/Sanitize.php on line 372``

This notice shows up in the logs every 5 minute. Haven't been able to find out what feed yet, but I'll let you know once I find out...

I'm using ``1.4 beta``
Haven't tested in ``1.3.1`` (yet), I'm about to migrate to simplepie...

-----------------------

```php
list($_, $data, $_) = explode($unique_tag, $data);
```
https://github.com/simplepie/simplepie/blob/master/library/SimplePie/Sanitize.php#L372

JanPetterMG:
**Here are the problematic URLs:** (the list will be updated as I discover them)
http://m.lubbockonline.com/taxonomy/term/240886/2/feed
http://podkast.nrk.no/program/ekko_-_et_aktuelt_samfunnsprogram.rss

droplister:
I am also getting this issue with some feeds. Here's some snippets from my error log. I am using Laravel.

```
[2016-05-10 21:54:55] local.ERROR: exception 'ErrorException' with message 'Undefined offset: 2' in /var/www/rssmixer.com/vendor/simplepie/simplepie/library/SimplePie/Sanitize.php:372
Stack trace:
#0 /var/www/rssmixer.com/vendor/simplepie/simplepie/library/SimplePie/Sanitize.php(372): Illuminate\Foundation\Bootstrap\HandleExceptions->handleError(8, 'Undefined offse...', '/var/www/rssmix...', 372, Array)
#1 /var/www/rssmixer.com/vendor/simplepie/simplepie/library/SimplePie.php(2118): SimplePie_Sanitize->sanitize('<p>&nbsp;</p>??...', 2, 'http://share.dm...')
#2 /var/www/rssmixer.com/vendor/simplepie/simplepie/library/SimplePie/Item.php(181): SimplePie->sanitize('<p>&nbsp;</p>??...', 2, 'http://share.dm...')
#3 /var/www/rssmixer.com/vendor/simplepie/simplepie/library/SimplePie/Item.php(323): SimplePie_Item->sanitize('<p>&nbsp;</p>??...', 2, 'http://share.dm...')
```

Although these probably aren't that helpful...

```
[...]

PR #442, opened 2016-04-25
Update version to 1.4.

mblaney:


mblaney:
SimplePie 1.4 is available! Thanks to everyone who has contributed since the last release.

skyzyx:
Wooooooo!!!

JanPetterMG:
Why isn't the phpDoc fixes in PR #440 merged? You really want to release 1.4.1 just for them?

And what about the recent discovered problems? #441 affecting multiple feeds at multiple pages, wouldn't it be fair to wait a few more days?

I can tell a fix is needed, no matter who's problem it is, an error is generated, due to badly written php code. Either we should check the input (_easy fix_), or fix the problem (_may be complicated, I don't know_)

I'll wait for 1.4.1, (or rolling back to ``1.3.1``) because there is no difference between the buggy ``1.4-beta`` and now released ``1.4``...

mblaney:
Hi @JanPetterMG, you did say you were going to keep adding to #440, so thought it best to hold off until you're done. Not that I really agree with the change, but if it works better for you that's fine.

#441 just generates a notice, but PR happily accepted.

mblaney:
@skyzyx do you know who has access to the simplepie.org website? Would be great to make some updates there too.

skyzyx:
Last I heard (≈5 years ago), @rmccue was working with someone on the WordPress team to take over hosting the site. He would know better than I would what happened after that.

rmccue:
@mblaney Sorry for the late update here; forgot to comment when I read it! Drop me an email at the address [on my site](https://rmccue.io/) under Contact and I can get you set up with everything :)

Yay for updates!

PR #445, opened 2016-05-15
Removed trailing and leading whitespaces

paolooo:
Removed trailing and leading whitespaces on XML feed 
content because dom/xml reader will throw an error is 
there's a space at the beginning.

See issue:  
http://screencast.com/t/rOuj1W3z


Alkarex:
This introduced a regression, breaking UTF-16LE and UTF-32BE encodings.
Fixed in https://github.com/simplepie/simplepie/pull/916

PR #462, opened 2016-08-09
Added PHP 7.1 for testing

JanPetterMG:
I'll see there is already commited patches for PHP 7.1, so why isn't it tested?


mblaney:
thanks @JanPetterMG

PR #463, opened 2016-08-11
PHPUnit dev-dependency

JanPetterMG:
Added PHPUnit as dev-dependency
- Version 4 required for PHP 5.3 support
- Version 6 is never going to work (when released) because of simplepies usage of deprecated class constructors, so we have to stop at 5.x


mblaney:
thanks again @JanPetterMG!

JanPetterMG:
You're welcome

PR #470, opened 2016-08-29
Use saveHTML(node) (was: Sanitize: superfluous entities)

Alkarex:
Re-use https://github.com/simplepie/simplepie/pull/388 instead of https://github.com/simplepie/simplepie/pull/378
Requires PHP 5.3.6, but fall-back with saveXML() for PHP 5.0.
https://github.com/simplepie/simplepie/issues/334
https://github.com/jagermesh/simplepie/commit/d91ae5e718a446652efd08cb42a622c93226d215

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -11,7 +11,9 @@ compatibility and standards compliance][what_is].
 
 Requirements
 ------------
-* PHP 5.3+
+* PHP 5.3.0+ (5.3.6+ recommended since SimplePie 1.4.3+)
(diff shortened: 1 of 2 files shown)


Alkarex:
@mblaney As agreed. I have also put a fall-back for PHP < 5.3.6, using `saveXML(node)`, which will likely produce slightly different results, but should probably be good enough (it would be nice to do some tests to see some actual differences).
I have not personally tested this patch yet on my systems.

PR #480, opened 2016-11-05
Use current HHVM 3.15.2 for testing

photodude:
The current Travis CI setup is testing against hhvm 3.6.6 since Travis CI container uses precise which is locked at hhvm 3.6.6 as the last hhvm build for that version. Using the non-container Trusty build allows testing current hhvm 3.15.2 (or newer depending on release schedule)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,18 @@
+sudo: false
 language: php
-php:
-  - 5.3
-  - 5.4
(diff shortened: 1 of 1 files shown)


mblaney:
thanks @photodude

photodude:
you are welcome @mblaney thanks for the merge

PR #489, opened 2016-11-22
Fix a PHP 7 deprecation message

brandonkelly:
As noted on PHP’s [Constructors and Destructors page](http://php.net/manual/en/language.oop5.decon.php), 

> Old style constructors are DEPRECATED in PHP 7.0, and will be removed in a future version. You should always use __construct() in new code.

PHP has supported __construct() since version 5 and SimplePie requires 5.3, so no reason not to change this now, and avoid deprecation errors when running PHP 7.

### DIFF
--- a/idn/idna_convert.class.php
+++ b/idn/idna_convert.class.php
@@ -91,7 +91,7 @@ class idna_convert
     var $_strict_mode    =  false;  // Behave strict or not
 
     // The constructor
-    function idna_convert($options = false)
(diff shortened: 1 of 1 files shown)


brandonkelly:
@mblaney Thanks for merging that! Any idea when the next release will be?

JanPetterMG:
Hopefully soon, am I right?
Just wait until 478 is closed, that shouldn't take too long...

PR #490, opened 2016-11-22
Fix #478

### LINKED ISSUE #478: get_content() behaviour and a potential improvement
Found a feed with an item structure like the following:

```
<description>
    <![CDATA[
        Here some description of the item
    ]]>
</description>
<content:encoded>
    <![CDATA[ ]]>
</content:encoded>
```

For this item, get_content() is returning the whitespace string inside `<content:encoded>` tag. I guess the publisher could do the things better, but should get_content() try to deal with this??

Below changes in `get_content()` solves the issue but not sure how far should SimplePie try to go to solve what seems to be a publisher mistake.

```
public function get_content($content_only = false)
{
    $atom10Content = $this->get_item_tags(SIMPLEPIE_NAMESPACE_ATOM_10, "content");
    $atom03Content = $this->get_item_tags(SIMPLEPIE_NAMESPACE_ATOM_03, "content");
    $rss10Content = $this->get_item_tags(SIMPLEPIE_NAMESPACE_RSS_10_MODULES_CONTENT, "encoded");
    $description = $this->get_description(true);
    if ($atom10Content && trim($atom10Content[0]["data"]))
    {
        return $this->sanitize($atom10Content[0]["data"], $this->registry->call("Misc", "atom_10_content_construct_type", array($atom10Content[0]["attribs"])), $this->get_base($atom10Content[0]));
    }
    elseif ($atom03Content && trim($atom03Content[0]["data"]))
    {
        return $this->sanitize($atom03Content[0]["data"], $this->registry->call("Misc", "atom_03_construct_type", array($atom03Content[0]["attribs"])), $this->get_base($atom03Content[0]));
    }
    elseif ($rss10Content && trim($rss10Content[0]["data"]))
    {
        return $this->sanitize($rss10Content[0]["data"], SIMPLEPIE_CONSTRUCT_HTML, $this->get_base($rss10Content[0]));
    }
    elseif (!$content_only)
    {
        return $description;
    }
    else
    {
        return null;
    }
}
```

JanPetterMG:
- Trim sanitized data
- Trim item data
- `get_content`: Override empty `<content:encoded>` tags with `<description>`
- `get_description`: Override empty `<description>` tags with `<content:encoded>`

Fixes both issues mentioned in #478, as well as a few others not mentioned yet, all related to trim.

_Hope I didn't violate the coding standards too much :p_

### DIFF
--- a/library/SimplePie/Item.php
+++ b/library/SimplePie/Item.php
@@ -326,57 +326,57 @@ class SimplePie_Item
 	 */
 	public function get_description($description_only = false)
 	{
-		if (($return = $this->get_item_tags(SIMPLEPIE_NAMESPACE_ATOM_10, 'summary')) &&
-			$return[0]['data'] &&
-			($return = $this->sanitize($return[0]['data'], $this->registry->call('Misc', 'atom_10_construct_type', array($return[0]['attribs'])), $this->get_base($return[0])))
+		if (($tag = $this->get_item_tags(SIMPLEPIE_NAMESPACE_ATOM_10, 'summary')) &&
+			$tag[0]['data'] &&
+			($return = $this->sanitize($tag[0]['data'], $this->registry->call('Misc', 'atom_10_construct_type', array($tag[0]['attribs'])), $this->get_base($tag[0])))
 		) {
 			return $return;
 		}
-		elseif (($return = $this->get_item_tags(SIMPLEPIE_NAMESPACE_ATOM_03, 'summary')) &&
-			$return[0]['data'] &&
-			($return = $this->sanitize($return[0]['data'], $this->registry->call('Misc', 'atom_03_construct_type', array($return[0]['attribs'])), $this->get_base($return[0])))
+		elseif (($tag = $this->get_item_tags(SIMPLEPIE_NAMESPACE_ATOM_03, 'summary')) &&
+			$tag[0]['data'] &&
+			($return = $this->sanitize($tag[0]['data'], $this->registry->call('Misc', 'atom_03_construct_type', array($tag[0]['attribs'])), $this->get_base($tag[0])))
 		) {
 			return $return;
 		}
-		elseif (($return = $this->get_item_tags(SIMPLEPIE_NAMESPACE_RSS_10, 'description')) &&
-			$return[0]['data'] &&
-			($return = $this->sanitize($return[0]['data'], SIMPLEPIE_CONSTRUCT_MAYBE_HTML, $this->get_base($return[0])))
+		elseif (($tag = $this->get_item_tags(SIMPLEPIE_NAMESPACE_RSS_10, 'description')) &&
+			$tag[0]['data'] &&
+			($return = $this->sanitize($tag[0]['data'], SIMPLEPIE_CONSTRUCT_MAYBE_HTML, $this->get_base($tag[0])))
 		) {
 			return $return;
 		}
-		elseif (($return = $this->get_item_tags(SIMPLEPIE_NAMESPACE_RSS_20, 'description')) &&
-			$return[0]['data'] &&
-			($return = $this->sanitize($return[0]['data'], SIMPLEPIE_CONSTRUCT_HTML, $this->get_base($return[0])))
+		elseif (($tag = $this->get_item_tags(SIMPLEPIE_NAMESPACE_RSS_20, 'description')) &&
+			$tag[0]['data'] &&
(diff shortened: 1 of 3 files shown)


mblaney:
Hi @JanPetterMG this won't help, the original problem was that data was just a space, which is still truthy. You need to check what `sanitize` is returning before deciding to continue or not. Also closing bracket on the next line is a pretty big departure from the current coding standard. And finally `trim` is called at the start of `sanitize`, you shouldn't need to call it again at the end.

JanPetterMG:
Not sure what you mean, there is no space anymore, my tests passes, and the problem is eliminated!
Could you please share your test file?

The coding standards, yes it's a little bit less readable due to the if else complexity, but it's PSR-2 witch is pretty common these days... Submitting an patch to revert to the old school style.

library/SimplePie/Sanitize.php
```diff
@@ -404,7 +404,7 @@ public function sanitize($data, $type, $base = '')
 				$data = $this->registry->call('Misc', 'change_encoding', array($data, 'UTF-8', $this->output_encoding));
 			}
 		}
-		return $data;
+		return trim($data);
```
JanPetterMG: @mblaney I won't remove trim here, as it's needed!
Obviously we could call `trim` each time `$data` is updated, but is this really worth it, compared to calling it once here at the end?

library/SimplePie/Item.php
```diff
@@ -90,7 +90,26 @@ class SimplePie_Item
 	public function __construct($feed, $data)
 	{
 		$this->feed = $feed;
-		$this->data = $data;
+		$this->data = $this->array_map_recursive('trim', $data);
```
JanPetterMG: @mblaney The space you we're talking about, it's non-existing because of this line...

mblaney: I should've mentioned that too. The constructor is the wrong place to be manipulating data like this. It doesn't logically make sense to say that data should be trimmed here, as there is nothing inherently wrong with having white space.

mblaney:
Hi @JanPetterMG I've added a test that will fail to show you what I mean: https://github.com/mblaney/simplepie/commit/fef8a16c81ba985c9e97de80ac0b8c695ef4da0c

The idea being that if we're going to return the sanitized version anyway, we should at least check that *it* is not empty, otherwise it's no different from the space issue.

Regarding coding standards, it's not about what is more common but what is consistent. The only way the SimplePie coding standard is going to change is if someone changes the whole code base... and believe me, I've considered doing that just to remove the tabs ;-)

library/SimplePie/Item.php
JanPetterMG: **Where does it make sense to return an untrimmed whitespace?**
Sorry if I don't see it, but I don't think an trim would have any negative effect on any of the affected functions. On the positive side, we won't need to trim anything multiple times, just because we're using the data in multiple places... #performance

Here's the full list:
`get_authors`, `get_base`, `get_categories`, `get_content`, `get_contributors`, `get_copyright`, `get_date`, `get_description`, `get_enclosures`, `get_id`, `get_item_tags`, `get_latitude`, `get_links`, `get_longitude`, `get_source`, `get_thumbnail`, `get_title`, `get_updated_date`.

JanPetterMG:
I'll see, I didn't test with empty tags inline, but that part should now be covered too.

I've still left the `$return[0]['data']` in the `if else` statement, for performance reasons, no need to try `sanitize` if `get_item_tags` already is returning empty data inside the array.
Again, the initial `array_map_recursive('trim', $data)` saves us for additional wasted cpu cycles. Both compared to checking `sanitize` when the content already is empty, and calling `trim` in each `if else` statement.

All I'm doing is turning the tile upside down :-)

mblaney:
@JanPetterMG great, I'm glad you understand what I was saying about `sanitize` now. :-)

Sorry but I still disagree with most of these changes though. I don't believe `$data` should be modified unless there's a good reason for it, and I just don't see one. We shouldn't be adding code unless it's required, and "performance reasons" is too subjective. Changes should be functional, and even better, write a test to prove it's required.

JanPetterMG:
@mblaney What if you fix this? You know obviously better.
This PR fixes both #478 and #493, so I'll be expecting an patch for both of them!

mblaney:
@JanPetterMG #494 fixes #478, I don't agree with the issues you've raised in #493, as I've said before I don't believe there's anything inherently wrong with untrimmed data. If there are real-world problems (like #478) I'm happy to look at it then.

JanPetterMG:
@mblaney Then you'll have to take a look again...
Simplepie is basically inserting an whitespace that isn't present in the original string.

Issue #422, opened 2015-12-31
Why is <dc:subject> treated as category?


Lukas0907:
Hi,

I was wondering why `<dc:subject>` is treated as a category by the `get_categories()` methods.

A subject and a category seem to be different things to me.

Cheers,

Lukas

JanPetterMG:
From the Dublin Core spec: http://dublincore.org/documents/2012/06/14/dcmi-terms/?v=elements#subject

> **Definition:** The topic of the resource.
> **Comment:** Typically, the subject will be represented using keywords, key phrases, or classification codes. Recommended best practice is to use a controlled vocabulary.

After some research, this table seems to be the best explanation: http://www.feedsweep.com/ShowArticle.aspx?ID=38

| Filter Item | Atom | RSS (2.0, 0.91, 0.92) | RDF/RSS (1.0, 0.90) |
| ----- | ----- | ----- | ----- |
| FeedTags | /atom/feed/category | /rss/channel/category /rss/channel/dc:subject | /rdf:RDF/channel/dc:subjec |

-----

**Some personal thoughts:**
I'll agree, there is too many categories used in Atom/RSS feeds worldwide, a lot of them seems to be too unique (eg. reverse lookup returning 1 item, for a timeline of 3 years, when filtering by url host).

I haven't done any research on `dc:subjec` specifically, but I guess it has an small to medium impact on the results above.
The biggest problem as I see it, is the webpages doing the following:
- tagging no categories at all
- tagging a hardcoded list of categories
- tagging categories that's not really an category (eg. author, agency, it's own webpage name, etc).

JanPetterMG:
This issue is resolved in version `1.5`.
Close it whoever has the ability to do so.

PR #492, opened 2016-11-25
Get category type

JanPetterMG:
`SimplePie_Category->get_type()` will return either `category` for `<category>`, or `subject` for `<dc:subject>`.

See also #422

Bonus feature, `SimplePie_Category->get_label($strict=false)` now accepts bool. `true` to return label only, or default `false` for fallback to `get_term()` in cases where the label doesn't exist.

_Once merged, we'll have to bump to v1.5_

### DIFF
--- a/library/SimplePie/Category.php
+++ b/library/SimplePie/Category.php
@@ -56,7 +56,7 @@ class SimplePie_Category
 	/**
 	 * Category identifier
 	 *
-	 * @var string
+	 * @var string|null
 	 * @see get_term
 	 */
 	var $term;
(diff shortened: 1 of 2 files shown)


mblaney:
Hi @JanPetterMG this looks good, thanks. Just one closing bracket on the following line, and no need to split the `if` statement into multiple lines... so if you can change `get_label` to:
`if ($this->label === null && $strict === false)`

I would be happy to merge :-)

JanPetterMG:
Fixed.
It's less readable, not in this case, but in more complex if else statements it can be an huge advantage.

Is it possible to release an 1.4.3 with important fixes before merging this? That will give me/us a little more time to focusing on other minor changes...

PR #494, opened 2016-11-27
Check result of sanitize before returning in get_content and get_description

mblaney:


JanPetterMG:
So you're just ignoring my suggestion to add `!empty($tags[0]['data'])` in the `if else` statements?

mblaney:
yes because `sanitize` already handles this case very quickly.

PR #504, opened 2017-02-27
Declare `$enable_exceptions` class property

### LINKED ISSUE #330: Undefined property: SimplePie::$enable_exceptions on line 1982 in /simplepie/library/SimplePie.php
The enable_exceptions property is not defined in SimplePie, so an error is triggered when this property is used on line 1982:

``` php
    public function sanitize($data, $type, $base = '')
    {
        try
        {
            return $this->sanitize->sanitize($data, $type, $base);
        }
        catch (SimplePie_Exception $e)
        {
            if (!$this->enable_exceptions)
            {
```

kevincupp:
Fixes #330


mblaney:
thanks @kevincupp

PR #509, opened 2017-04-09
Allow force-get supplied item ID

JanPetterMG:
Allow forcing `$item->get_id()` to return supplied ID, instead of generating one. If no ID is supplied, it will return `null`.

**The issue:**
If an item gets updated (eg. a news article), and a item ID isn't supplied, a new unique ID will __always__ be generated, causing duplicates, even if it has the same permalink or even title...

**Bonus changes/fixes:**
- Previously, if an invalid or not callable function name was supplied, it was replaced by `md5` without warning, now it produces an `E_USER_WARNING` first...
- The `$fn` parameter, was missing from the PHPdoc block, now fixed.

### DIFF
--- a/library/SimplePie/Item.php
+++ b/library/SimplePie/Item.php
@@ -206,9 +206,10 @@ class SimplePie_Item
 	 *
 	 * @since Beta 2
 	 * @param boolean $hash Should we force using a hash instead of the supplied ID?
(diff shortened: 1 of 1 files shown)


mblaney:
thanks @JanPetterMG

PR #512, opened 2017-04-18
Feeds behind proxies

### LINKED ISSUE #498: Feed not retrieved when behind a proxy
Hello,

We got an issue trying to retrieve some feeds from behind a proxy.

For some URLs, there are no problems; but for others, the feed cannot be retrieved. Looking more deeper into the problem, I think the way that is used to "calculate" the number of header in `File.php` is in cause.

On some proxy responses, we got 2 "HTTP 200" in the header, but with no redirect; so the explode in https://github.com/simplepie/simplepie/blob/master/library/SimplePie/File.php#L139 will give unexpected results.

As an example on GitHub's blog atom feed; header is:
```
HTTP/1.1 200 OK
Server: GitHub.com
Date: Thu, 19 Jan 2017 12:49:26 GMT
Content-Type: application/atom+xml; charset=utf-8
Transfer-Encoding: chunked
Status: 200 OK
```

Behind a proxy, it becomes:
```
HTTP/1.0 200 Connection established


HTTP/1.1 200 OK
Server: GitHub.com
Date: Thu, 19 Jan 2017 12:47:49 GMT
Content-Type: application/atom+xml; charset=utf-8
Transfer-Encoding: chunked
Status: 200 OK
```

With that one; `SimplePie_HTTP_Parser` will fail. I've quick fixed that while testing with the following patch:
```diff
diff -up simplepie/library/SimplePie/File.php.patch simplepie/library/SimplePie/File.php
--- simplepie/library/SimplePie/File.php.patch	2017-01-19 13:08:23.003879459 +0100
+++ simplepie/library/SimplePie/File.php	2017-01-19 13:09:06.968313793 +0100
@@ -136,8 +136,8 @@ class SimplePie_File
 						$this->url = $info['url'];
 					}
 					curl_close($fp);
-					$this->headers = explode("\r\n\r\n", $this->headers, $info['redirect_count'] + 1);
-					$this->headers = array_pop($this->headers);
+					$this->headers = explode("HTTP/", $this->headers);
+					$this->headers = "HTTP/" . array_pop($this->headers);
 					$parser = new SimplePie_HTTP_Parser($this->headers);
 					if ($parser->parse())
 					{
```

But that sounds a bit ugly to me... Any ideas? I wonder if this could be handled on the parser side; but redirections handling would be handled other there as well, and not in `File`.

PS: to get the issue from unit tests
```diff
diff --git a/tests/HTTPParserTest.php b/tests/HTTPParserTest.php
index 283aa72..82dc27e 100755
--- a/tests/HTTPParserTest.php
+++ b/tests/HTTPParserTest.php
@@ -80,4 +80,19 @@ class HTTPParserTest extends PHPUnit_Framework_TestCase
                $this->assertEquals($expected, $parser->body);
 
        }
-}
\ No newline at end of file
+
+       /**
+        * @dataProvider chunkedProvider
+        */
+       public function testChunkedProxy($data, $expected)
+       {
+               $data = "HTTP/1.0 200 Connection established\r\n\r\nHTTP/1.1 200 OK\r\nContent-Type: text/plain\r\nTransfer-Encoding: chunked\r\n\r\n" . $data;
+               $parser = new SimplePie_HTTP_Parser($data);
+               $this->assertTrue($parser->parse());
+               $this->assertEquals(1.1, $parser->http_version);
+               $this->assertEquals(200, $parser->status_code);
+               $this->assertEquals('OK', $parser->reason);
+               $this->assertEquals(array('content-type' => 'text/plain'), $parser->headers);
+               $this->assertEquals($expected, $parser->body);
+       }
+}
```

trasher:
Fixes #498

### DIFF
--- a/library/SimplePie/File.php
+++ b/library/SimplePie/File.php
@@ -136,8 +136,7 @@ class SimplePie_File
 						$this->url = $info['url'];
 					}
 					curl_close($fp);
-					$this->headers = explode("\r\n\r\n", $this->headers, $info['redirect_count'] + 1);
-					$this->headers = array_pop($this->headers);
+					$this->headers = SimplePie_HTTP_Parser::prepareHeaders($this->headers, $info['redirect_count'] + 1);
 					$parser = new SimplePie_HTTP_Parser($this->headers);
 					if ($parser->parse())
 					{
--- a/library/SimplePie/HTTP/Parser.php
+++ b/library/SimplePie/HTTP/Parser.php
@@ -496,4 +496,22 @@ class SimplePie_HTTP_Parser
 			}
 		}
 	}
+
+	/**
+	 * Prepare headers (take care of proxies headers)
+	 *
+	 * @param string  $headers Raw headers
+	 * @param integer $count   Redirection count. Default to 1.
+	 *
+	 * @return string
+	 */
+	static public function prepareHeaders($headers, $count = 1)
+	{
+		$data = explode("\r\n\r\n", $headers, $count);
+		$data = array_pop($data);
+		if (false !== stripos($data, "HTTP/1.0 200 Connection established\r\n\r\n")) {
+			$data = str_ireplace("HTTP/1.0 200 Connection established\r\n\r\n", '', $data);
+		}
+		return $data;
+	}
 }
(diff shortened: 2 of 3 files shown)


mblaney:
looks good, thanks @trasher

dpovshed:
Please see #548

PR #525, opened 2017-06-14
Updating favicon url in `get_favicon` function

abhij89007:
Updating outdated url http://g.etfv.co/ with google's url to fetch favicon. This will return the correct favicon.


mblaney:
thanks @abhij89007!

Issue #528, opened 2017-07-13
Clean up the demo examples


anewuser:
"BBC Arabic", "BBC Russia", "Windows Vista Blog" and likely other examples from this page are down: https://www.simplepie.org/demo/

anewuser:
@HardeepAsrani @mblaney  You left out a non-linked ".us".

mblaney:
thanks @anewuser I've fixed it on the server, @HardeepAsrani you can file a PR if you like?

PR #535, opened 2017-08-06
Removed/updated feed URLs

HardeepAsrani:
Removed the ones that weren't available, and changed the ones that changed domains(engadged, for example).

Fixes: https://github.com/simplepie/simplepie/issues/528


mblaney:
thanks @HardeepAsrani I've pushed your changes to the demo page.

HardeepAsrani:
I use WordPress and we have a plugin which uses this library. Glad to be of any help. Keep the great work up. :)

PR #536, opened 2017-08-07
Switch from regex to xpath for microformats discovery.

mblaney:


mblaney:
Let me know if anyone has feedback about this change, otherwise I'll just keep merging things. I've got a few more to add yet.

PR #548, opened 2017-10-16
Improvement to #512:  https://github.com/simplepie/simplepie/pull/512

dpovshed:
While worked with SQuid proxies, I found myself in the similar situation. The only difference is that I've gotten from proxy reponse with HTTP 1.1, not HTTP 1.0.

The pull request managed this situation as well. I just extended tests and code of the function SimplePie_HTTP_Parser::prepareHeaders in the simplest way, to not fall into overengineering.


mblaney:
thanks @dpovshed

PR #557, opened 2017-11-05
Add toJson() function support

JefferyHus:
In my case I have an API that calls some feed url then returns the results found, the front-end is VueJS base, so basically JSON is preferred as a response, why this function? because instead of doing too much on PHP side a `for loop` to build a complete list of items then convert it into JSON consumes time and memory, while it is so simple to do it in the Core class of `SimplePie_Item`.

### DIFF
--- a/library/SimplePie/Item.php
+++ b/library/SimplePie/Item.php
@@ -117,6 +117,16 @@ class SimplePie_Item
 	}
 
 	/**
+	 * Get a json string of the item
+	 *
+	 * @return string
+	 */
+	public function toJson()
+	{
+		return json_encode(array_shift($this->data['child']));
+	}
+	
+	/**
 	 * Remove items that link back to this before destroying this object
 	 */
 	public function __destruct()


mblaney:
Hi @JefferyHus thanks for the patch! Is there a reason you want to use array_shift? I don't think anyone calling toJson() would expect the function to remove data.

JefferyHus:
`array_shift` because the element containing data has an empty key:
```php
{
  "": {//data here}
}
```
So I had to shift the element tor return a full JSON object that contains the data everyone is looking for.

mblaney:
The key here is defined in the SimplePie namespace, I don't see a good reason to remove it from the item. There are lots of ways to access the data you're after via the namespace, see other functions in Item.php for examples.

JefferyHus:
Okay I will give it a shot and back to you.

JefferyHus:
will this be any good?

mblaney:
sorry I'm not following what you're saying.

PR #561, opened 2017-11-23
Use force_fsockopen & curl_options for autodiscovery.

MrDys:
The ability to set curl options was recently added, but autodiscovery was not making use of these options. This PR passes in the $force_fsockopen and $curl_options parameters to Locator so that when Locator makes the File instantiation to check validity of the feeds, those options will be taken into account.

### DIFF
--- a/library/SimplePie.php
+++ b/library/SimplePie.php
@@ -1640,7 +1640,7 @@ class SimplePie
 		if (!$this->force_feed)
 		{
 			// Check if the supplied URL is a feed, if it isn't, look for it.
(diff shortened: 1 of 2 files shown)


mblaney:
looks good @MrDys is there a reason not to use the extra parameters elsewhere in Locator.php where create File is called?

MrDys:
@mblaney Oh, nope. Just an oversight on my part. Updated the PR with those params.

mblaney:
great, thanks!

PR #565, opened 2017-12-16
Support PHPUnit 6

carusogabriel:
I've added `PHPUnit 6` support. I just need to bump `PHPUnit 4` to [`4.8.35`](https://github.com/sebastianbergmann/phpunit/blob/4.8/ChangeLog-4.8.md#4835---2017-02-06) and `PHPUnit 5` to [`5.4.3`](https://github.com/sebastianbergmann/phpunit/blob/5.4/ChangeLog-5.4.md#543---2016-06-09) to support the new `PHPUnit\Framework\TestCase` :man_technologist: 

The failing tests are because we are not installing anything on `Travis CI`, neither update or use `vendor/bin/phpunit`. Can I fix it in another PR?

PR #567, opened 2017-12-19
Fix TravisCI by running composer install

carusogabriel:
As suggested in #565


mblaney:
cool, thanks again @carusogabriel

PR #576, opened 2018-02-22
Fix tavis tests

peter279k:
# Changed log

- Nowadays, it should let the version PHP 7+ pass the tests during doing Travis CI build.
- In the ```PHP 7.x```, the ```PHPUnit_Framework_Error_Notice``` class is not found.
To be compatiable with the PHP 5.x, I use the ```class_alias``` to let ```PHPUnit_Framework_Error_Notice``` alias the ```PHPUnit\Framework\Error\Notice```.
- Consider the old tests and I sperate two different ```phpunit.xml.dist``` setting files.
It can see more details about the ```composer.json``` and ```.travis.yml```.
- Fix the assertions when executing Travis CI build.
In some test cases, the port number in url can identify the difference.
I notice that the ```substr``` function is diffreent in PHP ```5.x``` and PHP ```7.x```.
This [link](http://php.net/manual/en/function.substr.php) can see more details about the ```substr``` difference in different PHP version.
In PHP ```7+```, If the start character length is equal to string length, it will return empty string.
But in PHP ```5.x``` it will return ```boolean(false)```.
It happens on line ```916``` of ```IRI.php```.


mblaney:
Hi @peter279k thanks for this, I'm just trying to understand why you don't want to run the old tests in PHP 7?

peter279k:
Hi @mblaney, thank you for your reply.
After I add the old tests for PHP 7+, it has some errors.
I fix the error and test well.
Here is the lists about the error:

- The constructor must use ```__construct``` since ```PHP 7```. Otherwise it will be output the deprecated message.
- The PHP Internal ```usort``` function are different in ```PHP 5.x``` and ```PHP 7.x```.
Here are the links that can look at this details.
> * [link1](https://stackoverflow.com/questions/37941808/php-7-usort-adds-equal-items-to-end-of-array-where-in-php-5-it-adds-to-the-begin)
> * [link2](https://stackoverflow.com/questions/44542089/usort-difference-php7-1-vs-php5-6)

skyzyx:
@peter279k wrote:
> I notice that the substr function is diffreent in PHP 5.x and PHP 7.x. This link can see more details about the substr difference in different PHP version.

I think that this can be pretty easily patched in the source code, no?

> In the PHP 7.x, the `PHPUnit_Framework_Error_Notice` class is not found. To be compatiable with the PHP 5.x, I use the class_alias to let `PHPUnit_Framework_Error_Notice` alias the `PHPUnit\Framework\Error\Notice`.

I think this change happened in PHPUnit 6.x, which still supports PHP 5.5/5.6. I think it would be wise to fix the tests themselves to use the newer namespace instead of leveraging `class_alias`.

> The constructor must use `__construct` since PHP 7. Otherwise it will be output the deprecated message.

This change should have happened _ages_ ago. Again, let's fix the code itself.

> The PHP Internal `usort` function are different in PHP 5.x and PHP 7.x. Here are the links that can look at this details.

No, it's not that `usort` changed its behavior, so much as `return -1;` triggers _unknown behavior_. We just need to use better-defined behavior.

.travis.yml
```diff
@@ -1,23 +1,24 @@
-sudo: false
 language: php
+
+php:
+    - 5.4
+    - 5.5
+    - 5.6
```
skyzyx: We should remove 5.4 and 5.5 because they are _dead_.

.travis.yml
```diff
@@ -1,23 +1,24 @@
...
-    - php: 7.1
-    - php: hhvm
-      sudo: true
-      dist: trusty
-      group: edge # until the next Travis CI update
-  allow_failures:
-    - php: hhvm
-    - php: 7.0
-    - php: 7.1
+    - php: 5.3
```
skyzyx: Again, we should skip this version because it is dead.

.travis.yml
```diff
@@ -1,23 +1,24 @@
...
+      dist: precise
+
 before_script:
-    - travis_retry composer install --no-interaction --prefer-dist
+    - travis_retry composer install
```
skyzyx: This was better the first time.

composer.json
```diff
@@ -36,5 +36,11 @@
 		"psr-0": {
 			"SimplePie": "library"
 		}
-	}
+    },
+    "config": {
+        "bin-dir": "bin"
```
skyzyx: This is a big change. We shouldn't do it unless @mblaney (the maintainer) is OK with it, and any dependent scripts have been updated.

library/SimplePie.php
```diff
@@ -3146,6 +3146,12 @@ public static function sort_items($a, $b)
 	{
 		$a_date = $a->get_date('U');
 		$b_date = $b->get_date('U');
+		if(version_compare(phpversion(), '7.0.0', '>=')) {
```
skyzyx: Remove the `version_compare()`. The code you have for PHP 7 is actually right code for all platforms.

phpunit.xml.old.dist
```diff
@@ -0,0 +1,21 @@
...
+
+	<filter>
+		<blacklist>
+			<directory suffix=".php">tests</directory>
+		</blacklist>
+		<whitelist>
+			<directory suffix=".php">SimplePie</directory>
+		</whitelist>
+	</filter>
+</phpunit>
```
skyzyx: I need more proof that these files need to be different before accepting this.

tests/IRITest.php
```diff
@@ -41,9 +41,16 @@
...
+
 class IRITest extends PHPUnit\Framework\TestCase
 {
+    public static function setUpBeforeClass()
+    {
+        if(class_exists('PHPUnit_Framework_Error_Notice')) {
+            class_alias('PHPUnit_Framework_Error_Notice', 'PHPUnit\Framework\Error\Notice');
+        }
+    }
+
```
skyzyx: Again, I think we should just update PHPUnit to 6.x (which still supports PHP 5.6) and update all of the classname references.

phpunit.xml.old.dist
peter279k: This file I will be removed and use the ```phpunit.xml.dist```.

tests/IRITest.php
peter279k: No, the PHPUnit ```6.x``` cannot support the PHP 5.6.
If we want to support the PHP 5.6, we have to use the PHPUnit 5 and 6.
See more details about the PHPUnit [doc](https://phpunit.de/).

library/SimplePie.php
peter279k: This block should be existed because it separates the PHP 5.6 and PHP 7.
And it also prevents the unexpected ```usort``` behavior.

.travis.yml
peter279k: I will remove these dead PHP versions.

.travis.yml
peter279k: I will remove these PHP versions.

peter279k:
Hi @skyzyx, thank you for your reply.
The ```substr``` issue that I have solved and set the patch.
They're worked fine for the PHP 5.6 and PHP 7.
However, the ```usort``` cannot be good.
So it's necessary to seperate the two versions for using ```version_compare```.
If we want to prevent the unexpected ```usort``` behavior, we can use customize sorting approach.
These two issues are very hard to trace because of the existed legacy code.
I hope you can understand my delay.

composer.json
peter279k: All right. I can remove this and use the previous approach.

.travis.yml
peter279k: I think ```--prefer-dist``` should be removed because we have to check the dependency problem.

library/SimplePie.php
mblaney: why do you want to separate 5.6 & 7?
what is the unexpected `usort` behaviour?

mblaney:
Hi @peter279k thanks again, this is looking better.

Lets sort out the `usort` problems you're seeing, I'm happy to help.

Also just in general, if you can try not to touch files just to make white space changes the diff would be cleaner to read.

peter279k:
Hi @mblaney, thank you for your reply.
How about use the ```php_cs``` to set PSR-2 to set the white space problem?

library/SimplePie.php
mblaney: yes so it's the test using identical timestamps that expects an unreasonable value that's causing the problem here. I would just update tests/oldtests/first_item_title/SPtests/bugs/179.0.10.php

mblaney:
if you like, just keep the PR's separate :-)

mblaney:
Hi @peter279k just checking in to see if you're still interested in working on this? I think it's a good change, and with a couple more improvements could be merged.

peter279k:
Hi @mblaney, thank you for your reply.
I have commented the message after owner's request changes.
Just wait for the reply.

mblaney:
ok so just to summarise where we're up to, I suggested updating the problematic test, are you happy to do that? I agree with your handling of PHPUnit here, @skyzyx have you got any more thoughts on that one?

peter279k:
Hi @mblaney, do you mean the problematic test is about the ```tests/oldtests/first_item_title/SPtests/bugs/179.0.10.php```?

Thanks.

mblaney:
yes and on second thought I would just delete the test. I can't find what bug it's referring to, and giving the entry a unique timestamp just to pass the test would make it too similar to 179.0.10.1.php.

peter279k:
Hi @mblaney, you can check this Travis job [log](https://travis-ci.org/peter279k/simplepie/jobs/344634849).
Then take a look at the failure parts. Here is the failure message.

```
There were 3 failures:
1) IRITest::testStringNormalization with data set #9 ('http://example.com:', 'http://example.com/')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'http://example.com/'
+'http://example.com:0/'
/home/travis/build/peter279k/simplepie/tests/IRITest.php:314
2) IRITest::testStringNormalization with data set #16 ('https://example.com:', 'https://example.com/')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'https://example.com/'
+'https://example.com:0/'
/home/travis/build/peter279k/simplepie/tests/IRITest.php:314
3) IRITest::testStringNormalization with data set #27 ('http://[0:0:0:0:0:0:0:1]:', 'http://[::1]/')
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'http://[::1]/'
+'http://[::1]:0/'
/home/travis/build/peter279k/simplepie/tests/IRITest.php:314
```

mblaney:
why are you worried about those failures, I can see that you fixed them the next day?

peter279k:
I'm not worried about that and I just present these logs to refer the related bugs.
And it's about the line 3149 in ```library/SimplePie.php```.

mblaney:
I'm not following. There are no issues with `sort_items`, you just need to remove the test.

I don't see any other problems with this pull request, though it would still be good to hear from @skyzyx on the changes he requested.

mblaney:
Hi @peter279k I've removed the problematic test myself, so if you want to revert some of your white space only changes this pull request will be good to merge.

Thanks for your work getting tests happening for PHP 7. I'm not sure why there have been so many communications issues with this pull request, the basic change looks pretty good to me. I'll give you and @skyzyx a week or so to discuss this further if you like, and will otherwise merge after that.

library/SimplePie.php
mblaney: Hi @peter279k I was looking at merging your PR but noticed this change hasn't been reverted yet. Can you do that first please?

peter279k: Thank you for your reply, @mblaney.
I have removed this.

Thanks.

mblaney:
great, thanks again for the update @peter279k!

PR #589, opened 2018-07-24
Support for Redis password and database

pierrocknroll:
Hello,
A Redis URL can be like :
redis://[:password@]host[:port][/db-number]

For now, you only support host and port options. 

Thanks


mblaney:
thanks @pierrocknroll

pierrocknroll:
You're welcome @mblaney 
It was very easy actually.

Can you please make a release will all bugfixes since Nov 17 ?
Thanks

Issue #443, opened 2016-04-26
PHP 7 support


JanPetterMG:
Is PHP 7 support realistic?
http://php.net/supported-versions.php

There is currently 4 errors, 3 of them most likely due to the same port-number bug...
https://travis-ci.org/simplepie/simplepie 

---

_The coding standards used are way outdated, and with files counting 3.000+ lines, isn't making the job of fixing it easier either. Normally I'm trying to contribute instead of posting issues, but that's not realistic in this case... sorry!_

_**2.0:** Start the refactoring process, update the standards, replace old bunks of code with external maintained repositories, less code, fewer bugs, better support. Then it would be easier for everyone, also it will making sure this repository doesn't get out of date! Because who are using php 5.3 today??_

shaildyp:
PHP7 works for me but the the encoding parser keeps failing. I have both mbstring & iconv available.

`PHP Notice: The data could not be converted to UTF-8. You MUST have either the iconv or mbstring extension installed. Upgrading to PHP 5.x (which includes iconv) is highly recommended.`

idealley:
it also works for me with php 7 on vagrant/homestead

thinkingmedia:
I'm going to be running it on 7. If I see anything that needs fixing. I'll make a PR for it.

JanPetterMG:
Is backwards compatibility required at all cost? (there was an unspecified comment about this before 1.4 was released).
Is a ``2.0-dev`` branch possible?

3 of the issues are probably due to bugs in the [SimplePie_IRI class](https://github.com/simplepie/simplepie/blob/master/library/SimplePie/IRI.php). `1262 lines (1182 sloc)  28.1 KB`

I _might_ rewrite it, but then it has to be from scratch !!! The current class is very complex, are using outdated standards, and could be simplified a lot.

For example, the `replace_invalid_with_pct_encoding()` function with it's 275 lines, can be replaced with just 2 lines, still doing the exact same job... Solution: ``preg_replace`` + ``array`` + ``rawurlencode``

mblaney:
I think keeping backwards compatibility was suggested because SimplePie is included in Wordpress. It should be kept unless there's a really good reason not too, if that ever happens I would be open to a 2.0-dev branch, maintained in parallel.

JanPetterMG:
@shaildyp I also get that error, but on PHP 5.6.20, even if I have ``UConverter``, ``iconv`` and ``mbstring`` installed, so it can't be PHP 7 specific...

[...]

PR #590, opened 2018-07-24
Composer: Added missing php module requirements

JanPetterMG:
Required:
- `PCRE`, `XML` and `XMLReader` - Simplepie will start complaining if missing...

Optional:
- `cURL` - If not, `fsockopen` will be used for downloads
- `iconv`, `intl` and `mbstring` - Character encoding support

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,12 +24,19 @@
 		}
 	],
 	"require": {
-		"php": ">=5.6.0"
+		"php": ">=5.6.0",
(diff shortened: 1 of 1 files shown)


JanPetterMG:
Not sure why no one has done this already...

Lots of people having issues when upgrading to PHP 7, because the modules have to be installed separately. See also #443

mblaney:
thanks @JanPetterMG

PR #604, opened 2019-02-01
PHP 7.4 compatibility patch.

Maikuolan:
There is a strong likelihood of is_writeable being deprecated when PHP 7.4
is released in the future. To ensure compatibility with PHP 7.4,
is_writable should be used instead.

See: https://wiki.php.net/rfc/deprecations_php_7_4


mblaney:
thanks @Maikuolan

Issue #609, opened 2019-03-22
Patch to support BusinessWire.com


escallonia:
Our processing of BusinessWire feeds broke around 2019/02/27. 

For example:
http://feeds.businesswire.com/BW/Electronic_Design_Automation-rss

The attached patch enabled me to resume processing BusinessWire. I submitted this as pull request #610 https://github.com/simplepie/simplepie/pull/610

[patch.txt](https://github.com/simplepie/simplepie/files/2998476/patch.txt)

mblaney:
Hi @escallonia did you try asking them to fix it?

escallonia:
> 
> 
> Hi @escallonia did you try asking them to fix it?

Will do and update here. First I resolved the outage in my app!

escallonia:
The is in the response headers from BusinessWire.com:
`Content-Type: application/rss+xml, text/xml; charset=UTF-8`

1. Is this header "incorrect" and something they should change? What if they don't?
2. Does it harm SimplePie to support this header?
3. Or does it improve SimplePie to support this header?

skyzyx:
I feel like we need to draw a line somewhere, else SimplePie become more of a mess.

```http
Content-Type: application/rss+xml, text/xml; charset=UTF-8
```

This is completely invalid. We could shoe-horn it in, but what about the next set of completely broken headers? I worry about opening this door even more than it has been.

IMO, it's more appropriate for you to maintain your own patch, rather than having the development team accept + own this change upstream. I vote no.

escallonia:
Thanks for the comment. I'll try force-feed() and update you here. Feel free to let this sit until I update you.

escallonia:
@skyzyx and @mblaney and others, would you mind upvoting the answer [here](https://stackoverflow.com/questions/55346887/comma-separated-list-valid-in-content-type-header), or posting an answer? Or I will cite both this thread and stackoverflow when asking BusinessWire to fix it.

escallonia:
@mblaney, @skyzyx , I have started a dialog with BusinessWire on this topic and your feedback.

mblaney:
@escallonia thanks for reaching out to them. I agree with @skyzyx all for being flexible but this is taking things too far.

PR #610, opened 2019-03-22
Change to resume support of BusinessWire.com

escallonia:
Our processing of BusinessWire feeds broke around 2019/02/27.

For example:
http://feeds.businesswire.com/BW/Electronic_Design_Automation-rss

This change enabled me to resume processing BusinessWire.

### DIFF
--- a/library/SimplePie/Locator.php
+++ b/library/SimplePie/Locator.php
@@ -153,7 +153,8 @@ class SimplePie_Locator
 			$sniffed = $sniffer->get_type();
 			$mime_types = array('application/rss+xml', 'application/rdf+xml',
 			                    'text/rdf', 'application/atom+xml', 'text/xml',
-			                    'application/xml', 'application/x-rss+xml');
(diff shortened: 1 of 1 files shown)


library/SimplePie/Locator.php
```diff
@@ -153,7 +153,8 @@ public function is_feed($file, $check_html = false)
 			$sniffed = $sniffer->get_type();
 			$mime_types = array('application/rss+xml', 'application/rdf+xml',
 			                    'text/rdf', 'application/atom+xml', 'text/xml',
-			                    'application/xml', 'application/x-rss+xml');
+			                    'application/xml', 'application/x-rss+xml',
+					    'application/x-rss+xml, text/xml');
 			if ($check_html)
```
escallonia: BusinessWire apparently changed it's content-type header around 2019/02/27 to 'application/rss+xml, text/xml'.

escallonia:
As noted in #609, this patch is not needed. I used SimplePie::force_feed() instead.

PR #621, opened 2019-09-16
Locator: Try also links containing feed

jtojnar:
This is quite common. For example, on https://www.androidauthority.com/.


mblaney:
yes looks good, thanks @jtojnar

Issue #2, opened 2009-09-29
SimplePie::subscribe_url() should follow permanent redirects


rmccue:
_Originally reported by Ryan Parman as [issue 8](http://bugs.simplepie.org/issues/show/8)_

This is complicated because we mustn't cache temporary redirects, meaning it needs large changes in our HTTP client.

jtojnar:
Why was this closed? We encountered this issue in selfoss too: https://github.com/SSilence/selfoss/pull/1124

It can be fixed rather simply:

```patch
--- a/library/SimplePie.php
+++ b/library/SimplePie.php
@@ -1712,8 +1712,8 @@
 					}
 					$cache = $this->registry->call('Cache', 'get_handler', array($this->cache_location, call_user_func($this->cache_name_function, $file->url), 'spc'));
 				}
-				$this->feed_url = $file->url;
 			}
+			$this->feed_url = $file->url;
 			$locate = null;
 		}
 
```

jtojnar:
Here is MVE:

```
$feed = new \SimplePie();
$feed->set_feed_url('http://google.de/alerts/feeds/18120933104492874959/12861401000818240705');
$feed->enable_cache(false);
$feed->init();
$feedUrl = $feed->subscribe_url();
var_dump($feedUrl);
```

skyzyx:
It was closed because the issue had been open since 2009 with no movement. We decided not to kid ourselves about it ever happening.

If it’s important to you, please provide a PR _with tests_ to show that it actually does what it intends to do, and doesn’t break other functionality.

jtojnar:
Done #627

PR #627, opened 2019-10-04
Return effective feed URL when asking for non-permanent subscribe_url

jtojnar:
`subscribe_url` method supports two modes: permanent and non-permanent. And while the URL for the first one is obtained from the `File` after it has been fetched, discounting auto-discovery, the latter one remains the same from the time `set_feed_url` was called.

Because even the permanent mode concerns itself with redirection (although limited to a potential 301 as a first link in the chain), I would expect the non-permanent one to care doubly so.

This patch enables just that – after the feed is fetched, url property is updated, regardless of auto-discovery being done.

Fixes: #2

### DIFF
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -10,11 +10,26 @@ class MockSimplePie_File extends SimplePie_File
 	public function __construct($url)
 	{
 		$this->url = $url;
+		$this->permanent_url = $url;
 		$this->headers = array(
 			'content-type' => 'application/atom+xml'
 		);
 		$this->method = SIMPLEPIE_FILE_SOURCE_REMOTE;
 		$this->body = '<?xml version="1.0" encoding="utf-8"?><feed xmlns="http://www.w3.org/2005/Atom" />';
 		$this->status_code = 200;
+	}
+}
+
+/**
+ * Acts as a fake feed request that simulates first a permanent redirect from http:// URLs to https://,
+ * and then appends a date non-permanently.
+ */
+class MockSimplePie_RedirectingFile extends MockSimplePie_File
+{
+	public function __construct($url)
+	{
+		parent::__construct($url);
+		$this->permanent_url = str_replace('http://', 'https://', $url); // simulate 301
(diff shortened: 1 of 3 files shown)


mblaney:
hi @jtojnar thanks for this, I notice you've added `$suite->addTestSuite('SubscribeUrlTest');` to AllTests.php were you planning on adding another file?

jtojnar:
Right, forgot to stage a new file again. I really need to start using something like https://github.com/hyperfekt/git-stagelight

tests/bootstrap.php
```diff
@@ -18,3 +19,15 @@ public function __construct($url)
...
+
+/**
+ * Acts as a fake feed request that redirects to a different URL
+ */
+class MockSimplePie_RedirectingFile extends MockSimplePie_File
+{
+	public function __construct($url)
+	{
+		parent::__construct($url);
+		$this->url = 'https://example.com/new-feed/';
```
jtojnar: I do not really like hardcoding this in the Mock. The redirect URI should be set in test cases that require it but I do not see a way too do it since SimplePie requires mocking for even a simple things like this, rather than using DI, as is standard these days.

One possibility would be using anonymous classes from PHP 7, but we still need to support PHP 5.6, since it is maintained by Debian until June 2020.

* https://wiki.debian.org/PHP#Available_versions
* https://wiki.debian.org/LTS/Jessie

mblaney: could you just append a character to the url rather than replacing the whole string?

mblaney:
@jtojnar your new test raises a notice, would you mind looking in to that too?

jtojnar:
Fixed the `MockSimplePie_File` data to be valid.

tests/bootstrap.php
jtojnar: I just made it to change the URI scheme. It will make the difference harder to notice but at least it is not hard-coding stuff.

jtojnar:
Actually, https redirection would likely be permanent. The only case when I would use non-permanent redirect would be something like a feed of the day.

mblaney:
Nice one @jtojnar thanks for the extra work! :-)

PR #632, opened 2019-11-07
[BUGFIX] Add german short forms for weekdays and months

### LINKED ISSUE #631: [BUG] Missing german month short forms
#### Description

Parsing a date like `Do, 31 Okt 2019 23:52:59 -0400` makes the parser to return `NULL` instead of the correct timestamp.

#### Steps to Reproduce

```
$parse = new SimplePie_Parse_Date();
$result = $parse->parse('Fr, 11 Okt 2019 11:14:11 -0400');

var_dump($result);
```
If you change `Okt` to `Oct`, the expected timestamp is returned. 

#### Solution

The solution is to add the short forms of all months for the german language.

I will come up with a PR for that.

mschwemer:
The parsing of a date fails, if there are some shortforms missing and
used in the string, which mus be parsed. Adding these solves the issue.

Fixes #631


skyzyx:
So, this is not a bug. It's frustrating — sure — but not a bug.

The [RSS 2.0 specification](http://www.rssboard.org/rss-specification#ltpubdategtSubelementOfLtitemgt) defines dates as being [RFC 822](https://tools.ietf.org/html/rfc822#section-5), which defines **only** english values. Anything that is non-english is invalid.

That said, 15 years ago we made some exceptions for a couple of languages, so there is a precedent. @mblaney is the current maintainer of this project so I'll defer to his judgment on this PR, but I just wanted to clarify that this is not a bug in SimplePie — rather it's a bug in the feed.

Syntactically, this PR looks correct.

mblaney:
yes I agree @skyzyx, happy to improve the languages exception though thanks @mschwemer!

PR #633, opened 2019-11-22
Update SimplePie.php

diddledan:
Fix doctag type annotations to use valid types: `str` should be `string`.


mblaney:
thanks @diddledan

PR #643, opened 2020-03-28
Distinct cache for force_feed

Alkarex:
Ensure that feeds retrieved with the force_feed option do not share the
same cache than feeds retrieved without this option.
Otherwise, in a context with multiple clients, clients without the
force_feed option would have a behaviour (working or not) depending
entirely on what other clients with force_feed are doing. In other
words, it procudes a side-effect difficult to predict, which should be
avoided.

### DIFF
--- a/library/SimplePie.php
+++ b/library/SimplePie.php
@@ -1711,7 +1711,7 @@ class SimplePie
 					{
 						trigger_error("$this->cache_location is not writable. Make sure you've set the correct relative or absolute path, and that the location is server-writable.", E_USER_WARNING);
 					}
-					$url = $this->feed_url . ($this->force_feed ? '#force_feed' : '');
+					$url = $this->feed_url . '#force_feed';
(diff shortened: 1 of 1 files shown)


Alkarex:
P.S. Tested using the default cache system

Alkarex:
Travis is successful, but apparently the status on GitHub has not received the notification

mblaney:
thanks @Alkarex looks like a good idea. you don't need the check in `fetch_data()` though because you're already in the block `if (!$this->force_feed)`

Alkarex:
Here you are @mblaney

mblaney:
you don't need to modify `fetch_data()`at all.

Alkarex:
Ah, sorry, it is indeed a `if (!force_feed) {}` block and not a `if (force_feed) {}` ...
(Amusingly, it was precisely to avoid such errors that I initially kept the same exact syntax for both cases)
Fixed

PR #645, opened 2020-04-22
Remove old PHP directive conditionals

desrosj:
Since the minimum PHP version was raised to 5.6, there are two PHP directive checks that can be safely removed from conditional statements:

- [`zend.ze1_compatibility_mode`](https://www.php.net/manual/en/ini.core.php#ini.zend.ze1-compatibility-mode)
- [`safe_mode`](https://www.php.net/manual/en/ini.sect.safe-mode.php#ini.safe-mode)

Both of these directives were deprecated in PHP 5.3 and removed completely in 5.4, so the conditionals will always return `true`. I have also removed the `version_compare()` checks in the same `if` statements as they were checking for PHP < 5.6 (which is no longer supported).

- [X] Removed checks for removed PHP directives.
- [X] Removed version checks for PHP < 5.6 in the same control structures.

Removing these will prevent false positives when running code analysis tools such as PHP_CodeSniffer where rulesets flag instances of deprecated and removed directives (such as in the [PHPCompatibility](https://github.com/PHPCompatibility/PHPCompatibility) ruleset).

### DIFF
--- a/library/SimplePie.php
+++ b/library/SimplePie.php
@@ -706,7 +706,7 @@ class SimplePie
 	 */
 	public function __destruct()
 	{
-		if ((version_compare(PHP_VERSION, '5.6', '<') || !gc_enabled()) && !ini_get('zend.ze1_compatibility_mode'))
+		if (!gc_enabled())
 		{
 			if (!empty($this->data['items']))
 			{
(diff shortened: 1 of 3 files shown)


mblaney:
thanks @desrosj

PR #657, opened 2020-08-11
PHP 8.0: prevent ValueError for invalid encoding

jrfnl:
Check the encoding against the list of supported encodings before ever calling `mb_convert_encoding()`.

This prevents a fatal `ValueError` on PHP 8.0 when passing an invalid encoding and maintains the documented behaviour of returning `false` on an invalid encoding.

### DIFF
--- a/library/SimplePie/Misc.php
+++ b/library/SimplePie/Misc.php
@@ -364,11 +364,12 @@ class SimplePie_Misc
 		}
 
 		// Check that the encoding is supported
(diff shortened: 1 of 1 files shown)


jrfnl:
FYI: alternatively the order can remain the same, but then the first condition should be wrapped in a try/catch.

mblaney:
hi @jrfnl that's ok I think flipping it the way you have is fine. Thanks for all these changes they look good to me!

PR #661, opened 2020-10-07
Fix typo in MIME type in Content_Type_Sniffer

ericek111:
This surprised me. I may be misunderstanding something, but I believe `application/octet-stream` is the right MIME type for this function, seeing [there's no](https://www.iana.org/assignments/media-types/media-types.xhtml) `octect-stream`.


mblaney:
thanks @ericek111, looks like a typo. I noticed there's a few others in Misc.php but it's only used as a variable name.

JanPetterMG:
Is it possible to get a new release? Thanks!
This, as well as a few other minor changes, has happen since last release in may.

PR #670, opened 2021-01-29
Fix PHP8 crash due to insufficient isset test

Alkarex:
#fix https://github.com/FreshRSS/FreshRSS/issues/3401 (crash with PHP 8+)

`ceil()` crashes in PHP8+ in case of invalid input such as empty string.
`intval()` fixes the problem with almost identical behaviour than `ceil()` in PHP7- (except for floating point values)

#fix FreshRSS/FreshRSS#3401 (crash with PHP 8+)

Example with feed http://podcast.hr2.de/derTag/podcast.xml

```xml
<enclosure url="https://mp3podcasthr-a.akamaihd.net:443/mp3/podcast/derTag/derTag_20210129_87093232.mp3"
length="" type="audio/mpeg"/>
```

`isset("")` passes and then `ceil("")` crashes due to wrong type in PHP8+:

```
Uncaught TypeError: ceil(): Argument #1 ($num) must be of type
int|float, string given in ./SimplePie/SimplePie/Item.php:2871
```

### DIFF
--- a/library/SimplePie/Item.php
+++ b/library/SimplePie/Item.php
@@ -1803,7 +1803,7 @@ class SimplePie_Item
 							}
 							if (isset($content['attribs']['']['fileSize']))
 							{
-								$length = ceil($content['attribs']['']['fileSize']);
+								$length = intval($content['attribs']['']['fileSize']);
 							}
 							if (isset($content['attribs']['']['medium']))
 							{
@@ -2425,7 +2425,7 @@ class SimplePie_Item
 						}
 						if (isset($content['attribs']['']['fileSize']))
 						{
-							$length = ceil($content['attribs']['']['fileSize']);
+							$length = intval($content['attribs']['']['fileSize']);
 						}
 						if (isset($content['attribs']['']['medium']))
 						{
@@ -2790,7 +2790,7 @@ class SimplePie_Item
 					}
 					if (isset($link['attribs']['']['length']))
 					{
-						$length = ceil($link['attribs']['']['length']);
+						$length = intval($link['attribs']['']['length']);
 					}
 					if (isset($link['attribs']['']['title']))
 					{
@@ -2833,7 +2833,7 @@ class SimplePie_Item
 					}
 					if (isset($link['attribs']['']['length']))
 					{
-						$length = ceil($link['attribs']['']['length']);
+						$length = intval($link['attribs']['']['length']);
 					}
 
 					// Since we don't have group or content for these, we'll just pass the '*_parent' variables directly to the constructor
@@ -2868,7 +2868,7 @@ class SimplePie_Item
 					}
 					if (isset($enclosure[0]['attribs']['']['length']))
 					{
-						$length = ceil($enclosure[0]['attribs']['']['length']);
+						$length = intval($enclosure[0]['attribs']['']['length']);
 					}
 
 					// Since we don't have group or content for these, we'll just pass the '*_parent' variables directly to the constructor


library/SimplePie/Item.php
```diff
@@ -2862,11 +2862,11 @@ public function get_enclosures()
...
 					$url = $this->sanitize($enclosure[0]['attribs']['']['url'], SIMPLEPIE_CONSTRUCT_IRI, $this->get_base($enclosure[0]));
-					if (isset($enclosure[0]['attribs']['']['type']))
+					if (!empty($enclosure[0]['attribs']['']['type']))
 					{
 						$type = $this->sanitize($enclosure[0]['attribs']['']['type'], SIMPLEPIE_CONSTRUCT_TEXT);
 					}
-					if (isset($enclosure[0]['attribs']['']['length']))
+					if (!empty($enclosure[0]['attribs']['']['length']))
 					{
 						$length = ceil($enclosure[0]['attribs']['']['length']);
```
Alkarex, line 2871: Here was the actual crash with PHP8+

jtojnar: It will still crash when invalid `enclosure="foo"` attribute is provided. Perhaps rather than checking emptiness, we should check if the value is a number.

Alkarex: @jtojnar Right, this should do the job https://github.com/simplepie/simplepie/pull/670/commits/3000a70c2692988f67ea346e0dc3d4ae8289b724

Alkarex: `intval()` will cast any wrong value to 0

Alkarex:
The Travis integration does not work, but I passed the tests manually, and they are unchanged before/after my PR:

```
composer test
> phpunit
PHPUnit 6.5.14 by Sebastian Bergmann and contributors.

..............R..............................................   61 / 1314 (  4%)
.............................................................  122 / 1314 (  9%)
.............................................................  183 / 1314 ( 13%)
.............................................................  244 / 1314 ( 18%)
.............................................................  305 / 1314 ( 23%)
.............................................................  366 / 1314 ( 27%)
.............................................................  427 / 1314 ( 32%)
.............................................................  488 / 1314 ( 37%)
.............................................................  549 / 1314 ( 41%)
.............................................................  610 / 1314 ( 46%)
.............................................................  671 / 1314 ( 51%)
.............................................................  732 / 1314 ( 55%)
.............................................................  793 / 1314 ( 60%)
.............................................................  854 / 1314 ( 64%)
.............................................................  915 / 1314 ( 69%)
.............................................................  976 / 1314 ( 74%)
............................................................. 1037 / 1314 ( 78%)
............................................................. 1098 / 1314 ( 83%)
............................................................. 1159 / 1314 ( 88%)
............................................................. 1220 / 1314 ( 92%)
............................................................. 1281 / 1314 ( 97%)
.................................                             1314 / 1314 (100%)

Time: 1.45 seconds, Memory: 18.00MB

There was 1 risky test:

1) EncodingTest::test_convert_UTF8_uconverter with data set #0 (Binary String: 0xfeff221e, '∞', 'UTF-16')
This test did not perform any assertions

OK, but incomplete, skipped, or risky tests!
Tests: 1314, Assertions: 1422, Risky: 1.
```

library/SimplePie/Item.php
Alkarex: With that approach, the behaviour is unchanged compared to PHP7-

jtojnar: That should work (with slightly different semantics for numbers with decimal point) but now the `ceil` is not longer necessary.

Alkarex: Ah, right, on mobile now, but will simplify tomorrow

Alkarex: Otherwise `floatval()` could be an option to keep `ceil()`

Alkarex: `intval()` is probably better, since length and fileSize are both expressed in bytes anyway

mblaney:
thanks for fixing the `ceil` bug @Alkarex, but I don't agree with a blanket change from `isset` to `empty` because you're saying "0" isn't a valid string.

Alkarex:
@mblaney I kept `isset()` for href, so I did not change all of them. Which other attribute are you thinking about?

Alkarex:
@mblaney I have reverted to address only the specific problem of `ceil()`

PR #671, opened 2021-02-17
gitignore tests

Alkarex:
Ignore files created and downloaded by composer when running the tests locally.


jtojnar:
Sounds good but perhaps we should also remove `bin-dir` from `composer.json` and run `phpunit` from `vendor/bin` (composer does that [by default](https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands)).

PR #672, opened 2021-02-17
Reduce memory when parsing large feeds

Alkarex:
Upstream PR for https://github.com/FreshRSS/FreshRSS/pull/3416 (use case is 12MB+ feed)

Use the approach recommended by https://php.net/xml-parse#example-5983 for parsing documents that can potentially be large, because parsing a whole document in one go takes a lot of memory.

No change in parsing approach compared to now for feeds up to 1MB (i.e. most feeds are unchanged - in my list of 173 test feeds, only one is larger than 1MB). Larger feeds will be parsed in more than one iteration (no functional difference).

Using the php://temp as defined in https://php.net/wrappers.php fully in memory for feeds up to 2MB (by default) then using system's temp directory https://php.net/sys-get-temp-dir

There is a test for badly configured systems with an unwritable temp directory for which we only use php://memory (only in-memory even if it does not fit)

Credits to @Kiblyn11 for the idea and the original PR.

### DIFF
--- a/library/SimplePie/Parser.php
+++ b/library/SimplePie/Parser.php
@@ -164,12 +164,30 @@ class SimplePie_Parser
 			xml_set_element_handler($xml, 'tag_open', 'tag_close');
 
 			// Parse!
-			if (!xml_parse($xml, $data, true))
+			$wrapper = @is_writable(sys_get_temp_dir()) ? 'php://temp' : 'php://memory';
+			if (($stream = fopen($wrapper, 'r+')) &&
+				fwrite($stream, $data) &&
+				rewind($stream))
 			{
-				$this->error_code = xml_get_error_code($xml);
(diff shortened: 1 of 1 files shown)


library/SimplePie/Parser.php
```diff
@@ -164,12 +164,30 @@ public function parse(&$data, $encoding, $url = '')
 			xml_set_element_handler($xml, 'tag_open', 'tag_close');
 
 			// Parse!
-			if (!xml_parse($xml, $data, true))
+			$wrapper = @is_writable(sys_get_temp_dir()) ? 'php://temp' : 'php://memory';
```
Alkarex, line 167: Can be simplified by always using `php://memory` at the cost of not being able to parse just as large feeds

mblaney:
this looks great, thanks!

PR #673, opened 2021-03-20
PHP8 catch ValueError for loadHTML()

Alkarex:
Fix
```
PHP Fatal error:  Uncaught ValueError: DOMDocument::loadHTML(): Argument #1 ($source) must not be empty in /SimplePie/Locator.php:83
```

`set_error_handler` does not catch ValueError in PHP8.
The patch is compatible PHP5 and PHP7.

Fix downstream bug https://github.com/FreshRSS/FreshRSS/issues/3537
Downstream PR https://github.com/FreshRSS/FreshRSS/pull/3547

### DIFF
--- a/library/SimplePie/Locator.php
+++ b/library/SimplePie/Locator.php
@@ -75,12 +75,19 @@ class SimplePie_Locator
 		$this->force_fsockopen = $force_fsockopen;
 		$this->curl_options = $curl_options;
 
-		if (class_exists('DOMDocument'))
+		if (class_exists('DOMDocument') && $this->file->body != '')
 		{
 			$this->dom = new DOMDocument();
 
 			set_error_handler(array('SimplePie_Misc', 'silence_errors'));
-			$this->dom->loadHTML($this->file->body);
+			try
+			{
+				$this->dom->loadHTML($this->file->body);
+			}
+			catch (Throwable $ex)
+			{
+				$this->dom = null;
+			}
 			restore_error_handler();
 		}
 		else


library/SimplePie/Locator.php
```diff
@@ -75,12 +75,19 @@ public function __construct(SimplePie_File $file, $timeout = 10, $useragent = nu
 		$this->force_fsockopen = $force_fsockopen;
 		$this->curl_options = $curl_options;
 
-		if (class_exists('DOMDocument'))
+		if (class_exists('DOMDocument') && $this->file->body != '')
```
klausenbusk, line 78: Wouldn't it be better to use the [`empty()`](https://www.php.net/manual/en/function.empty.php) function? Can it be `null`?

library/SimplePie/Locator.php
```diff
@@ -75,12 +75,19 @@ public function __construct(SimplePie_File $file, $timeout = 10, $useragent = nu
...
 		{
 			$this->dom = new DOMDocument();
 
 			set_error_handler(array('SimplePie_Misc', 'silence_errors'));
-			$this->dom->loadHTML($this->file->body);
+			try
+			{
+				$this->dom->loadHTML($this->file->body);
+			}
+			catch (Throwable $ex)
```
klausenbusk, line 87: Is this needed or is it more better safe than sorry?

library/SimplePie/Locator.php
Alkarex: `loadHTML()` will raise an error if the input is null or empty. This test covers both.

library/SimplePie/Locator.php
Alkarex: There are potentially other errors than just empty input. For this specific bug, using either the null/empty test or the catch is sufficient. The `catch` is to more specifically extend the `set_error_handler`.

library/SimplePie/Locator.php
klausenbusk: Oh yeh, "type juggling".

PR #675, opened 2021-03-28
Prevent cache polution

Alkarex:
Retrieving a feed with distinct cURL parameters such as user-agent might give different responses, which should not share the same cache.

Follow-up and generalisation of https://github.com/simplepie/simplepie/pull/643

The code to get the cache filename is extracted in its own new function `SimplePie::get_cache_filename($url)` so that clients can easily get this information as well.

Downstream PR: https://github.com/FreshRSS/FreshRSS/pull/3502

### DIFF
--- a/library/SimplePie.php
+++ b/library/SimplePie.php
@@ -916,6 +916,39 @@ class SimplePie
 	}
 
 	/**
+	 * Return the filename (i.e. hash, without path and without extension) of the file to cache a given URL.
+	 * @param string $url The URL of the feed to be cached.
+	 * @return string A filename (i.e. hash, without path and without extension).
+	 */
+	public function get_cache_filename($url)
+	{
+		// Append custom parameters to the URL to avoid cache pollution in case of multiple calls with different parameters.
+		$url .= $this->force_feed ? '#force_feed' : '';
+		$options = array();
+		if ($this->timeout != 10)
+		{
+			$options[CURLOPT_TIMEOUT] = $this->timeout;
+		}
+		if ($this->useragent !== SIMPLEPIE_USERAGENT)
+		{
+			$options[CURLOPT_USERAGENT] = $this->useragent;
+		}
+		if (!empty($this->curl_options))
+		{
+			foreach ($this->curl_options as $k => $v)
+			{
+				$options[$k] = $v;
+			}
+		}
+		if (!empty($options))
+		{
+			ksort($options);
+			$url .= '#' . urlencode(var_export($options, true));
+		}
+		return call_user_func($this->cache_name_function, $url);
+	}
+
+	/**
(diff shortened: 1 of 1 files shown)


library/SimplePie.php
```diff
@@ -915,6 +915,39 @@ public function set_cache_location($location = './cache')
...
+	 * Return the filename (i.e. hash, without path and without extension) of the file to cache a given URL.
+	 * @param string $url The URL of the feed to be cached.
+	 * @return string A filename (i.e. hash, without path and without extension).
+	 */
+	public function get_cache_filename($url)
+	{
+		// Append custom parameters to the URL to avoid cache pollution in case of multiple calls with different parameters.
+		$url .= $this->force_feed ? '#force_feed' : '';
+		$options = array();
+		if ($this->timeout != 10)
```
Alkarex, line 928: Those tests are to keep the best possible compatibility with existing cache filenames: if not using custom cURL options, the cache filenames should be unchanged

library/SimplePie.php
```diff
@@ -915,6 +915,39 @@ public function set_cache_location($location = './cache')
...
+		{
+			foreach ($this->curl_options as $k => $v)
+			{
+				$options[$k] = $v;
+			}
+		}
+		if (!empty($options))
+		{
+			ksort($options);
+			$url .= '#' . urlencode(var_export($options, true));
```
Alkarex, line 946: The modified URL is still a valid URL, in case a custom `cache_name_function()` might have such an expectation

mblaney:
could this be optional? I don't feel that changing your agent necessarily means you want to cache both versions, but this is enforcing it. I think you will also need to move `force_feed` to your options array, because you could end up with two `#` segments in the url.

Alkarex:
1. Yes, we can make an option if you want. I am not sure what use-case would be addressed by such an option though. In any case, I think it would be more standard and less error prone to not share cache by default. As standard, HTTP caches are not supposed to be shared with different query parameters, and should even obey things such as `Vary` in the response. I you confirm that you rather want an option, and if yes what the default should be, I will edit accordingly.

2. Extra hashes are legal in an URL fragment, e.g. https://example.net/#one#two#three . See e.g. https://tools.ietf.org/html/rfc3986#appendix-B , any character is allowed after the first hash sign. The corresponding PHP function parses it just fine:

```php
<?php
$url = 'https://example.net/#one#two#three';
print_r(parse_url($url));
```

```
Array
(
    [scheme] => https
    [host] => example.net
    [path] => /
    [fragment] => one#two#three
)
```

mblaney:
ok fair enough, I was thinking caches might grow unexpectedly from this change but that's really only a problem if you're changing the agent options a lot. thanks for the update!

PR #676, opened 2021-03-28
Force HTTPS for selected domains

Alkarex:
Some feeds have embedded references using e.g. `//youtube.com/` or `http://youtube.com/` which are increasingly failing as HTTPS is required.

This PR allows setting a list of domains for which to force HTTPS with a new function `SimplePie::set_https_domains([])`

Downstream PR: https://github.com/FreshRSS/FreshRSS/pull/1087
https://github.com/FreshRSS/FreshRSS/pull/1098

### DIFF
--- a/library/SimplePie/Sanitize.php
+++ b/library/SimplePie/Sanitize.php
@@ -511,9 +511,9 @@ class SimplePie_Sanitize
 					if ($element->hasAttribute($attribute))
 					{
 						$value = $this->registry->call('Misc', 'absolutize_url', array($element->getAttribute($attribute), $this->base));
-						$value = $this->https_url($value);
-						if ($value)
+						if ($value !== false)
 						{
+							$value = $this->https_url($value);
 							$element->setAttribute($attribute, $value);
 						}
 					}
--- a/library/SimplePie/Sanitize.php
+++ b/library/SimplePie/Sanitize.php
@@ -72,6 +72,14 @@ class SimplePie_Sanitize
 	var $force_fsockopen = false;
 	var $replace_url_attributes = null;
 
+	/**
+	 * List of domains for which to force HTTPS.
+	 * @see SimplePie_Sanitize::set_https_domains()
+	 * Array is a tree split at DNS levels. Example:
+	 * array('biz' => true, 'com' => array('example' => true), 'net' => array('example') => array('www' => true))
+	 */
+	var $https_domains = array();
+
 	public function __construct()
 	{
 		// Set defaults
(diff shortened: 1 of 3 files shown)


library/SimplePie/Sanitize.php
```diff
@@ -241,6 +249,68 @@ public function set_url_replacements($element_attribute = null)
 		$this->replace_url_attributes = (array) $element_attribute;
 	}
 
+	/**
+	 * Set the list of domains for which to force HTTPS.
+	 * @see SimplePie_Misc::https_url()
+	 * Example array('biz', 'example.com', 'example.org', 'www.example.net');
+	 */
+	public function set_https_domains($domains)
```
Alkarex, line 257: The list of domains is transformed into a tree for fast lookup and to avoid ambiguities.

skyzyx:
Well I'm not opposed to this idea in general call mom I would hate to be on the hook for owning this list.

Is there anyway we could source a good list from somebody else who's already maintaining one? The EFF's HTTPS Everywhere extension comes to mind as a possible good source of data.

Assuming it's a JSON document, we could periodically fetch the list, transform it into a PHP lookup table, and ship a new release.￼￼￼

skyzyx:
My mistake; I was on my phone and didn't see that this is a bring-your-own-list situation.

I think this is a good idea.

Alkarex:
Not that it should be used, but just as an example, downstream, we use this list by default, customisable by end-users https://github.com/FreshRSS/FreshRSS/blob/edge/force-https.default.txt

Our two most important use-cases `dailymotion.com, youtube.com` are even set in the source-code at the moment but I will remove that if this PR is merged to keep the same code:

https://github.com/FreshRSS/FreshRSS/blob/634005de9a4b5e415ebf7c1106c769a0fbed5cfd/lib/SimplePie/SimplePie/Sanitize.php#L82

library/SimplePie/Sanitize.php
```diff
@@ -441,7 +511,8 @@ public function replace_urls($document, $tag, $attributes)
 					if ($element->hasAttribute($attribute))
 					{
 						$value = $this->registry->call('Misc', 'absolutize_url', array($element->getAttribute($attribute), $this->base));
-						if ($value !== false)
+						$value = $this->https_url($value);
```
mblaney, line 514: `absolutize_url` can return false, I don't think you should check the url in that case.

Alkarex: Fixed https://github.com/simplepie/simplepie/pull/676/commits/5bfec781a947654b4f3610690be1572389e93033

Issue #663, opened 2020-12-04
Make testsuite PHPUnit cross-version compatible


jrfnl:
####  Is your feature request related to a problem? Please describe.

At this moment the tests can not be run on PHP 8.0/nightly, due to the supported PHPUnit version being limited to PHPUnit 5.x and 6.x, which are not compatible with PHP 8.0.

As PHP 8.0 was released last week, it would be kind of nice to make sure the tests pass on PHP 8.0.

#### Describe the solution you'd like

Due to debilitating breaking changes in various PHPUnit versions between the currently supported versions and PHPUnit 9.3, which is the first version officially compatible with PHP 8.0, making the test suite cross-version compatible is not trivial.

The easiest way to do it, would be by either using [Symfony PHPUnit Bridge](https://github.com/symfony/phpunit-bridge) or the [PHPUnit Polyfills](https://github.com/Yoast/PHPUnit-Polyfills/) package.

#### Additional context

If the maintainers of this package are open to the idea of making the test suite PHPUnit cross-version compatible by using the PHPUnit Polyfills package, I'd be willing to create a PR for this.

jrfnl:
Anyone of the maintainers have an opinion on this ? The issue really becomes all the more urgent with the pletora of deprecation notices PHP 8.1 is throwing everywhere else. Would be nice to get some insight into the status of this repo ahead of the release.

With Travis dying an inglorious dead, I'd like to propose moving CI to GH Actions as well.

As I said above, I'm willing to put in the time & do the work, but would like the blessing of the maintainers for these proposals to prevent me wasting time.

/cc @mblaney @skyzyx

skyzyx:
@mblaney is the current maintainer. I just pop in from time to time these days to help out with the backlog.

My opinion is that:

1. This is a good idea.

2. We should err on the side of fixing it "the right way" instead of "the hacky way". I don't know _precisely_ what that means in this context, but it's meant as _general_ guidance for how I'd like to see us move forward.

3. We should drop support for EOL'd PHP versions because they're EOL. IIRC, 7.3 goes EOL in November/December. Too many people are interested in "making it work"; not enough people are interested in asking "_should_ we be doing this?" Using EOL'd PHP is a Bad Idea™.

4. If the _community_ wants to see support maintained for EOL'd versions, and these polyfills help with that, so be it.

That said, I'm an alumnus and not the current maintainer. @mblaney would be handling whatever merges and ongoing maintenance comes in. Therefore, it's ultimately his call.

[...]

PR #688, opened 2021-08-10
CI: switch to GH Actions

jrfnl:
Travis is dead, long live Travis!

This commit:
* Adds a GH Actions workflow for the CI checks which were previously run on Travis.
* Removes the, now redundant, `.travis.yml` configuration.
* Updates the `.gitattributes` file.

Other typical changes which I would usually make, do not apply to this repo.
* ~~Updates the "Build Status" badge in the Readme to use the results from the GH Actions runs.~~

Significant differences between this script and the original Travis script:
* I've not perpetuated the branch exception for the `one-dot-two` branch as that branch seems to be abandoned anyhow.
* Added testing against PHP 8.0 and PHP 8.1, with the test run against both being allowed to fail (for now).
    On PHP 8.0, the tests do actually run, but show a failure.
    On PHP 8.1, the tests do not run due to the PHPUnit version being used not being compatible with PHP 8.1 and as that version of PHPUnit is no longer supported, it won't be made compatible with PHP 8.1 anymore either.

Related #663


jrfnl:
FYI: to verify the build, you can look at the action run in my fork: https://github.com/jrfnl/simplepie/runs/3294162182?check_suite_focus=true

mblaney:
thanks @jrfnl, so you're allowing it to fail with the aim of fixing the failures next, is that correct?

jrfnl:
@mblaney Correct ✔️

PR #690, opened 2021-08-14
Tests: PHPUnit cross-version compatibility

### LINKED ISSUE #663: Make testsuite PHPUnit cross-version compatible

jrfnl:
This PR makes the test suite cross-version compatible with PHPUnit 5.x-9.x, which then allows the tests to be run on the full range of supported PHP versions, including PHP 8.1 (beta).

Note: this doesn't mean the tests will _pass_, but test failures will/should be addressed in separate PRs.

As this test suite uses a limited set of the PHPUnit functionality, not that much needed to be done and where necessary, I've just added a trait to an individual test file.

If in the future, you'd want the full range of polyfills to be available at all times to all tests, I'd recommend to add a generic `SimplePieTestCase` which either extends one of the Polyfill test cases or includes all the Polyfills traits. The actual test classes can then all extend from that `TestCase`. Or alternatively, all test classes could extend one of the Polyfill test cases instead of the PHPUnit base TestCase.

Fixes #663

### Tests: make config cross-version compatible

PHPUnit config file:

* Add `backupGlobals="false"`.
    The default value for this setting changed in PHPUnit 6 from `true` to `false`. By explicitly setting it to `false`, the behaviour for the tests will be consistent PHPUnit cross-version.
* Remove the `blacklist` element. This element hasn't been supported since PHPUnit 5.0.
* Point to the correct directory for the source code to allow for recording code coverage.
* Make sure that all src files are taken into consideration when calculating code coverage.
* Add XSD schema reference.
    Note: the config as-is will now validate for PHPUnit 4.4-9.2. For PHPUnit 9.3, the code coverage terminology has changed, though the "old" configuration is still supported.
    If needs be (PHPUnit 10), the config can be updated on the fly by using `--migrate-configuration`.

Other:
* Ignore a locally overloaded PHPUnit config file using the standard `phpunit.xml` file name.

### Composer: add dependency on the PHPUnit Polyfills package

* Adds a dev dependency to the `yoast/phpunit-polyfills` package.
* As that package already requires and manages the installable versions for PHPUnit, remove this as an explicit requirement from `require-dev` in favour of letting the PHPUnit Polyfills package manage the versions.
    This will update the supported PHPUnit versions from `~5.4.3 || ~6.5` to `^5.7.21 || ^6.0 || ^7.0 || ^8.0 || ^9.0`.
    Note: the supported versions for PHPUnit 5.x are very specific and restrictive as it specifically targets the versions in which the "forward compatible" alias files for the namespaced classes are available.

This new package adds the following features:
* Polyfills for various PHPUnit cross-version changes.
* Basic test case and test listener.

Refs:
* https://github.com/Yoast/PHPUnit-Polyfills/

Includes
* Adding the PHPUnit cache file, which is automatically created in PHPUnit 8 and 9, to the `.gitignore` file.
* Adding a `require_once` statement to load the PHPUnit Polyfills autoload file from the test bootstrap.

### Tests: switch out @ExpectedException annotations

These annotations are no longer supported in recent PHPUnit versions.

For PHP native errors/warnings/notices, the `TestCase::expectError()`, `TestCase::expectWarning()`, `TestCase::expectNotice()` and `TestCase::expectDeprecation()` methods (and variants) should be used.

For (custom) exceptions, the `TestCase::expectException()` method (and variants) should be used.

This also removes the need for the `setUpBeforeClass()` method in the `IRITest`, which only created an alias for a PHPUnit class.

The new methods and their variants are polyfilled via the PHPUnit Polyfill repo via the included trait.

### EncodingTest: rename a custom assertion

The method `assertEquals()` is a PHPUnit native method and if it is overloaded, the method signature is expected to be the same as the method signature in PHPUnit itself.

However, the method signature for this method has changed a couple of times across PHPUnit versions.

The simplest way to get round this is to rename the custom assertion to not overlap with a PHPUnit native method name.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ SimplePie.compiled.php
 bin/
 vendor/
 composer.lock
+phpunit.xml
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ bin/
 vendor/
 composer.lock
 phpunit.xml
+.phpunit.result.cache
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
 		"ext-xmlreader": "*"
 	},
 	"require-dev": {
-		"phpunit/phpunit": "~5.4.3 || ~6.5"
+		"yoast/phpunit-polyfills": "^1.0.1"
 	},
 	"suggest": {
 		"ext-curl": "",
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,21 +1,23 @@
 <?xml version="1.0" encoding="UTF-8"?>
-
-<phpunit colors="true" bootstrap="tests/bootstrap.php">
-	<testsuites>
+<phpunit
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/7.5/phpunit.xsd"
+    backupGlobals="false"
+    bootstrap="tests/bootstrap.php"
+    colors="true"
+    >
+    <testsuites>
         <testsuite name="SimplePie Test Suite">
             <directory>./tests</directory>
         </testsuite>
         <testsuite name="Old Tests">
-			<file>tests/oldtests.php</file>
+            <file>tests/oldtests.php</file>
         </testsuite>
     </testsuites>
 
     <filter>
-	    <blacklist>
-		    <directory suffix=".php">tests</directory>
-        </blacklist>
-        <whitelist>
-            <directory suffix=".php">SimplePie</directory>
+        <whitelist addUncoveredFilesFromWhitelist="true">
+            <directory suffix=".php">library</directory>
         </whitelist>
     </filter>
 </phpunit>
--- a/tests/CacheTest.php
+++ b/tests/CacheTest.php
@@ -41,6 +41,8 @@
  * @license http://www.opensource.org/licenses/bsd-license.php BSD License
  */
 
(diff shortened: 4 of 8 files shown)


jrfnl:
P.S.: I have a commit ready to address the test failure on PHP 8+, but it depends on this PR.

I can either pull it after this PR has been merged or add it to this PR, even though it is not part of the scope of this PR (PHPUnit cross-version compat).

mblaney:
looks good happy to merge the next PR separately

jrfnl:
@mblaney I've opened PR #691 to address that last test failure.

Issue #616, opened 2019-07-17
[BUG] Redis options not set - default to expire 0


greg606:
#### Description

When using Redis Cache there is no way to set redis cache expiry. 
Url params are not taken into account

#### Steps to Reproduce

```php
$remoteFeed = new SimplePie();
$remoteFeed->set_cache_location('redis://localhost:6379/?timeout=3600&prefix=sp_&dbIndex=0');
$remoteFeed->set_cache_duration(3601);
$remoteFeed->set_feed_url('any url');
$remoteFeed->init();
```
#### Expected Behavior

TTL of redis key not -1

#### Actual Behavior

TTL of redis key is -1

#### Possible Solutions
https://github.com/simplepie/simplepie/blob/11ddaed291ccd1b2628eef433a6bef58a1e4f920/library/SimplePie/Cache/Redis.php#L63
is always null 
also constructor has wrong phpdoc

#### Environment
dev-master
php-redis on linux mint, php 7.3

AnTheMaker:
This still isn't fixed yet, is it? I'm running into the same bug...

AnTheMaker:
Made a PR to fix this bug, hope it get's merged!

Art4:
With SimplePie 1.8.0 it will be possible to set Redis as a PSR-16 cache implementation and modify there the cache expiry.

PR #695, opened 2021-08-23
Fix Redis expiry (#616)

AnTheMaker:
This should fix a bug in the Redis cache-handler which caused the `timout` and `prefix` paramters to be ignored. Because of that you couldn't set a custom prefix for the Redis keys and all Redis data was stored indefinitely, without an expiry/timeout (Issue #616).

The correct way to use the Redis cache-handler (according to the code comments and documentation) is the following:
`redis://localhost:6379/?timeout=3600&prefix=sp_`

Until now however, the `timout` and `prefix` paramters were ignored. The Parameters already got parsed, so I just added two IFs to check if they are set and use them (or fall back to the defaults which were already there in the code if they aren't set).

If there are any questions, please let me know!
I hope this gets merged (or fixed differently) as I'd love to use the Redis cache-handler instead of the flat-file cache but I can't use it because of the missing expiry functionality (at least until now)!

### DIFF
--- a/library/SimplePie/Cache/Redis.php
+++ b/library/SimplePie/Cache/Redis.php
@@ -16,9 +16,11 @@
  *
  * Registered for URLs with the "redis" protocol
  *
- * For example, `redis://localhost:6379/?timeout=3600&prefix=sp_&dbIndex=0` will
+ * For example, `redis://localhost:6379/?timeout=3600&prefix=sp_` will
  * connect to redis on `localhost` on port 6379. All tables will be
- * prefixed with `simple_primary-` and data will expire after 3600 seconds
+ * prefixed with `sp_` and data will expire after 3600 seconds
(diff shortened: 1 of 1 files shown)


mblaney:
Hi @AnTheMaker you're right that those url parameters are being ignored, but passing `prefix` and `expire` in as the `$options` parameter should work? As it is your patch breaks that use by assigning an empty array to `$options`.

Art4:
Hey @AnTheMaker Since SimplePie 1.8.0 the `SimplePie\Cache\Redis` class is deprecated. You can workaround your issue by providing the Redis cache as PSR-16, e.g. using this adapter: https://github.com/php-cache/redis-adapter

@mblaney I think, this PR can be closed.

AnTheMaker:
Alright, thanks a lot for the info @Art4!

PR #698, opened 2021-10-11
fix: better deal with proxy returning proxy headers (in response to cURL's Proxy-Connection header)

shunf4:
Some HTTP proxy returns headers like this:

```
HTTP/1.1 200 Connection established\r\n
Connection: close\r\n
(Other headers...)
\r\n
(Headers and bodies of actual HTTP response)
```

So the code just detecting `"HTTP/1.0 200 Connection established\r\n\r\n"` won't work with these proxies.

A better way will be split the payload by `\r\n\r\n`, skip the first part and keep the remaining parts.


mblaney:
ok thanks @shunf4

Issue #669, opened 2021-01-25
[BUG] 


andrewpolemeni:
class-simplepie.php is causing the following issue:

[Mon Jan 25 13:38:14.133045 2021] [php7:warn] [pid 18879] [client 13.66.139.39:26051] PHP Warning: preg_match() expects parameter 2 to be string, array given in /wp-includes/class-simplepie.php on line 2620

The code in bold is causing the issue since it's getting an array instead of a string.
`
if (isset($this->data['headers']['link']) && preg_match('/<([^>]+)>; rel='.preg_quote($rel).'/', **$this->data['headers']['link']**, $match))
        {   
            return array($match[1]);
        }
        else if (isset($this->data['links'][$rel]))
        {
            return $this->data['links'][$rel];
        }
        return null;
    }
`

andrewpolemeni:
@skyzyx Why did you remove your assignment? Will this bug get fixed?

skyzyx:
I removed my assignment because I haven't been an active developer on this project in 10 years. :)

SimplePie has been in _low-power mode_ for quite some time. If the community wants to step-up with a patch, cool. But I will not be the person to write that patch, unfortunately.

skyzyx:
```php
$this->data['headers']['link']
```

The [Link header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link) has always been an array, even it commonly only has a single element. If it were me, knowing that it _should_ always be an array, I would:

1. Check the type.
2. Massage any scalar value into an array with a single element.
3. Update this code to work with `[]string` instead of expecting a `string`.

voegelas:
The feed https://blog.jetbrains.com/idea/feed/ sends two Link headers and triggers this bug. SimplePie is used from DokuWiki, which doesn't seem to use the Link headers. I have replaced `isset` by `is_string` as a quick fix.

The command `curl -v https://blog.jetbrains.com/idea/feed/ >/dev/null` outputs the headers below.

    Link: <https://blog.jetbrains.com/wp-json/>; rel="https://api.w.org/"
    Link: </>; rel=shortlink

PR #700, opened 2021-11-11
Handle multiple Link headers

voegelas:
This pull request updates the method `get_links` in `library/SimplePie.php` as suggested in #669. The updated code handles both single and multiple Link headers.


mblaney:
thanks @voegelas looks good

PR #709, opened 2021-12-20
Update CHANGELOG.md, follow keepachangelog format

Art4:
I've noted that the `CHANGELOG.md` was not updated since 1.5.5, so I updated it. And I formated the file to follow the [keepachangelog.com](https://keepachangelog.com/en/1.0.0/) format.

I couldn't find a note if this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html) as I assumed in the CHANGELOG.md, so please correct me if I'm wrong.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,55 @@
-## 1.5.5 (May 1, 2020)
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+## [1.5.7](https://github.com/simplepie/simplepie/compare/1.5.6...1.5.7) - 2021-12-19
+
+* Fix PHP8 crash due to insufficient isset test by @Alkarex in #670
+* gitignore tests by @Alkarex in #671
+* Reduce memory when parsing large feeds by @Alkarex in #672
+* PHP8 catch ValueError for loadHTML() by @Alkarex in #673
+* Provide access to HTTP status code by @Alkarex in #674
+* Fix wrong type hint by @Alkarex in #678
+* Force HTTPS for selected domains by @Alkarex in #676
+* Prevent cache polution by @Alkarex in #675
+* Fix typo in comment by @Alkarex in #680
+* Remove HTTP credentials in HTTP Referer by @Alkarex in #681
+* CI: switch to GH Actions by @jrfnl in #688
+* PHP 8.1: fix "passing null to non-nullable" deprecation notice by @jrfnl in #689
+* Tests: PHPUnit cross-version compatibility by @jrfnl in #690
+* Tests: use strict assertions by @jrfnl in #692
+* CacheTest: handle different exceptions PHP cross-version by @jrfnl in #691
+* GH Actions: don't allow builds to fail on PHP 8.0 or 8.1 by @jrfnl in #693
+* Tests: use the correct parameter order by @jrfnl in #694
+* PHPUnit: update configuration by @jrfnl in #696
+* fix: better deal with proxy returning proxy headers (in response to cURL's Proxy-Connection header) by @shunf4 in #698
+* Handle multiple Link headers by @voegelas in #700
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,26 +1,71 @@
-## 1.5.5 (May 1, 2020)
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
```
jtojnar, line 5: Is not the main point of Keep a Changelog that changes are split per type?

CHANGELOG.md
```diff
@@ -1,26 +1,71 @@
...
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+## [1.5.7](https://github.com/simplepie/simplepie/compare/1.5.6...1.5.7) - 2021-12-19
+
+* Fix PHP8 crash due to insufficient isset test by @Alkarex in #670
+* gitignore tests by @Alkarex in #671
```
jtojnar, line 13: IMO, the changelog should be curated and not contain inconsequential stuff like CI changes. Author names are probably too noisy for it.

CHANGELOG.md
```diff
@@ -1,26 +1,71 @@
...
+* Reduce memory when parsing large feeds by @Alkarex in #672
```
jtojnar, line 14: It would be useful to use issue links like in the previous releases since GitHub does not linkify shortlinks nor would they be clickable when viewing the changelog locally.

CHANGELOG.md
Art4: The main point of Keep a Changelog is to **keep** a changlog for humans. 😊 No one has to follow the format ideologically. Nearly all rules are recommendations, that have stood the test of time.

If the CHANGELOG.md is not up-to-date or not written at all, then it does not help anyone. Splitting them into groups could be a next step if the maintainers gives the format a 👍.

CHANGELOG.md
Art4: The format stated that Changelogs should be written for humans. For me it looks like the maintainers decided in the past to keep those changes (like changes to travis or composer) in the Changelog, so did I. But removing or rewriting those changes could be done in this or an next PR. This is up to the maintainers.

CHANGELOG.md
Art4: You are right, older entries are also linked to the issues. Thank you. I will fix this. 👍

Art4: I've added links to the issues.

CHANGELOG.md
jtojnar: I fully agree, I am just not sure if we should say it is based on Keep a Changelog when it does not follow one of its main recommendations.

CHANGELOG.md
jtojnar: Yeah, I interpret being written for humans as reducing noise since humans do not deal with it very well. Similarly, reader probably does not care who contributed each feature so I would suggest moving acknowledgements into a paragraph at the end – there we can thank the contributors while keeping the list of changes clean. But yeah, this is largely up to maintainers and their preferences.

CHANGELOG.md
Art4: I think following 5 of the 7 Guiding Principles, and using rfc3339 for dates is a good start. 😊 Splitting the changes into groups requires more time in each issue and before I do that work, I would like to know if that would be accepted by the maintainers. The same goes for marking deprecations or breaking changes.

Art4:
Some thoughts on backwards compatability.

A [comment from 2017](https://github.com/simplepie/simplepie/issues/443#issuecomment-310963472) by @skyzyx suggests that this project adheres to Semver and that "we should be responsible about messaging BC breaks". (But I don't see that confirmed or documented anywhere.)

The `CHANGELOG.md` and adhering Semver solve both of these problems. In the CHANGELOG.md deprecations can be documented. And breaking changes are only introduced via new major version and are documented in the CHANGELOG.md.

The first step would be for @mblaney to determine if SemVer is being adhering. The `CHANGELOG.md` file would be a good place to state that. Personally, I would be very happy to see this happen.

mblaney:
thanks @Art4 happy for you to keep the changelog updated

PR #716, opened 2022-01-07
Add audio, video @src elements/attribute for URL resolution

rdalverny:
Independently from enclosures, it happens that audio/video contents are also embedded through `<audio src="" />` and `<video src="" />` tags.
Those deserve their source URLs to be properly resolved too. 😛 

However I'm not 100% aware of the history of SimplePie, so there may be a reason those were left out (other than they're just rather recent elements). Thanks!

* [x] Tests ok
* [x] Linter ok
* (couldn't find a Makefile though as it's hinted in the PR notes)

### DIFF
--- a/tests/SanitizeTest.php
+++ b/tests/SanitizeTest.php
@@ -66,7 +66,53 @@ EOT
 		);
 	}
 
-    public function testSanitizeURLResolution()
+
+    public function sanitizeURLProvider()
+    {
+        return array(
+            'simple absolute valid a href, resolved' => array(
+                '<a href="/path/to/doc">link</a>',
+                '<a href="http://example.com/path/to/doc">link</a>'
+            ),
+            'image valid fully qualified src, no change expected' => array(
+                '<img src="http://2.example.com/image.jpg">',
+                '<img src="http://2.example.com/image.jpg">'
+            ),
+            'image valid relative src, resolved' => array(
+                '<img src="image.jpg">',
+                '<img src="http://example.com/image.jpg">'
+            ),
+            'image valid absolute src, resolved' => array(
+                '<img src="/image.jpg">',
+                '<img src="http://example.com/image.jpg">'
+            ),
+            'audio relative src, resolved, fixed' => array(
+                '<audio src="a.mp3" />',
+                '<audio src="http://example.com/a.mp3" preload="none"></audio>'
+            ),
+            'audio absolute source src path, resolved, fixed' => array(
+                '<audio><source src="/a/b.wav" /></audio>',
+                '<audio preload="none"><source src="http://example.com/a/b.wav"></audio>'
+            ),
+            'audio with alternative source src absolute paths, resolved, fixed' => array(
+                '<audio><source src="a/b.wav" /><source src="/c/d.mp3" /></audio>',
+                '<audio preload="none"><source src="http://example.com/a/b.wav"><source src="http://example.com/c/d.mp3"></audio>'
+            ),
+            'video src relative, resolved, fixed' => array(
+                '<video src="./b.mpeg" />',
+                '<video src="http://example.com/b.mpeg" preload="none"></video>'
+            ),
+            'video with alternative source src, resolved, fixed' => array(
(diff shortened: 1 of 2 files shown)


skyzyx:
I seem to recall that it was because the spec was _a lot_ more complex, and we never got around to it.

Can you please cite which part of which spec you're trying to implement here? Looking at the diff, this appears to be a _very_ incomplete patch for adding spec-compliant support for these elements.

rdalverny:
Thanks for the fast reply! :+1: And good point. The spec indeed is a bit more complex than I thought with this first change.

I tried to look for other closed similar issues for reference but couldn't find one.

To clarify: the goal with this PR: improve the support for `<audio>` and `<video>` elements URL resolution, so that audio/video media directly in the HTML may be previewed/played properly when relative URLs are used to reference the media file.

Relevant specs for [\<audio\>](https://html.spec.whatwg.org/multipage/media.html#the-audio-element), [\<video\>](https://html.spec.whatwg.org/multipage/media.html#the-video-element), [\<picture\>](https://html.spec.whatwg.org/multipage/embedded-content.html#the-picture-element) and [\<source\>](https://html.spec.whatwg.org/multipage/embedded-content.html#the-source-element).

My understanding:

1. `<audio>` element uses either the `src` attribute, either sub-elements `<source src="">`;
2. `<video>` element likewise (there's also a `poster` attribute that would also need resolution);
3. `<picture>` element uses `<source srcset="">`;
4. in all cases, we want the URL resolution to happen for `src` and `poster` attributes;
5. so "luckily" `<picture>` usage doesn't overlap with in this instance and we could let it aside here.

I also pushed tests & complements to cover these other attribute and element. I'm not sure if that's all spec-compliant, but it's a bit more complete, better illustrates the case and at least adds some support for it.

The bit I am not satisfied with however:

* is the cleanup/removal (in last commit) of the `</source>` close tag which is automatically added when [`$data = trim($document->saveHTML($div));` is run](https://github.com/simplepie/simplepie/blob/1aec297145a150f627e6ed82b8b084e0c933def8/library/SimplePie/Sanitize.php#L439) - perhaps because DOMDocument does not handle well HTML5?
  It works, but...
* another option would be to use `saveXML()` instead, but that breaks at least one other test (`ItemTest::testItemWithEmptyContent`). This might not be too hard to fix but I'm not certain it wouldn't have other side effects;
* another option?

Thanks again for your feedback!

mblaney:
@rdalverny not sure why your new tests are failing for older versions of php if you want to have a look at that?

tests/SanitizeTest.php
```diff
@@ -65,4 +65,45 @@ public function testSanitize()
...
+                => 'audio 3 <audio preload="none"><source src="http://example.com/a/b.wav"><source src="http://example.com/c/d.mp3"></audio>',
+            'video 1 <video src="./b.mpeg" />'
+                => 'video 1 <video src="http://example.com/b.mpeg" preload="none"></video>',
+            'video 2 <video><source src="a/b.mpeg"></video>'
+                => 'video 2 <video preload="none"><source src="http://example.com/a/b.mpeg"></video>',
+            'video 3 <video><source src="a/b.mpeg" /><source src="c/d.mov"></video>'
+                => 'video 3 <video preload="none"><source src="http://example.com/a/b.mpeg"><source src="http://example.com/c/d.mov"></video>',
+        );
+
+        foreach ($tests as $input => $expected)
```
jtojnar, line 104: Might be better to replace this by data provider to make these independent:

https://phpunit.readthedocs.io/en/9.5/writing-tests-for-phpunit.html#writing-tests-for-phpunit-data-providers

rdalverny:
Thanks to both of you.
This also fixes the tests (I ran them on 5.6 locally after this update, with correct behaviour).
I'm not sure why it failed, but suspect it has to do with the test strings with free text unwrapped before the HTML? (then handled differently depending on the versions)

mblaney:
looks good to me, @skyzyx did you have any more feedback?

mblaney:
this seems ok to me, so unless there's any follow up will merge this too

mblaney:
thanks @rdalverny merging now

PR #717, opened 2022-01-27
added: Sanitize::rename_attributes()

math-GH:
I want to bring a new method that we implemented into Simplepie of FreshRSS: https://github.com/FreshRSS/FreshRSS/pull/4175

The use case:
Simplepie can clean the HTML with `strip_attributes()`, but in some cases it would be very useful to keep the information of some attributes under another attribute name, f.e. `id` and `class`
The new `rename_attributes()` methods enables it. The `id` and `class` attributes could be renamed to `data-sanitized-id` / `data-sanitized-class`.

The benefit: when HTML/XML/RSS is included into an existing HTML DOM the imported HTML/XML/RSS could have id/class attributes that are already defined in the DOM, but with another meaning. Rename this attributes brings id/class out of standard CSS behavior but keeps the information. It is still possible to style it with `div[data-sanitized-class*="CLASSNAME"]`

It would be a pleasure to extend Simplepie with this little improvement.

### DIFF
--- a/library/SimplePie.php
+++ b/library/SimplePie.php
@@ -651,7 +651,7 @@ class SimplePie
 	public $strip_htmltags = array('base', 'blink', 'body', 'doctype', 'embed', 'font', 'form', 'frame', 'frameset', 'html', 'iframe', 'input', 'marquee', 'meta', 'noscript', 'object', 'param', 'script', 'style');
 
 	/**
-	 * @var array Stores the default tags to be stripped by rename_attributes().
+	 * @var array Stores the default attributes to be renamed by rename_attributes().
 	 * @see SimplePie::rename_attributes()
 	 * @access private
 	 */
--- a/library/SimplePie.php
+++ b/library/SimplePie.php
@@ -651,6 +651,13 @@ class SimplePie
 	public $strip_htmltags = array('base', 'blink', 'body', 'doctype', 'embed', 'font', 'form', 'frame', 'frameset', 'html', 'iframe', 'input', 'marquee', 'meta', 'noscript', 'object', 'param', 'script', 'style');
 
 	/**
+	 * @var array Stores the default tags to be stripped by rename_attributes().
+	 * @see SimplePie::rename_attributes()
+	 * @access private
+	 */
+	public $rename_attributes = array();
+
+	/**
 	 * @var bool Should we throw exceptions, or use the old-style error property?
 	 * @access private
 	 */
@@ -1216,6 +1223,15 @@ class SimplePie
 		{
 			$this->sanitize->encode_instead_of_strip($tags);
 		}
+	}
+
+	public function rename_attributes($tags = '', $encode = null)
+	{
+		if ($tags === '')
+		{
+			$tags = $this->rename_attributes;
+		}
+		$this->sanitize->rename_attributes($tags);
 	}
 
 	public function encode_instead_of_strip($enable = true)
--- a/library/SimplePie.php
+++ b/library/SimplePie.php
@@ -1225,18 +1225,18 @@ class SimplePie
 		}
 	}
 
-	public function rename_attributes($tags = '', $encode = null)
-	{
-		if ($tags === '')
-		{
-			$tags = $this->rename_attributes;
-		}
-		$this->sanitize->rename_attributes($tags);
-	}
-
 	public function encode_instead_of_strip($enable = true)
 	{
 		$this->sanitize->encode_instead_of_strip($enable);
+	}
+
+	public function rename_attributes($attribs = '')
+	{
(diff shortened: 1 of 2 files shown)


library/SimplePie.php
```diff
@@ -650,6 +650,13 @@ class SimplePie
 	 */
 	public $strip_htmltags = array('base', 'blink', 'body', 'doctype', 'embed', 'font', 'form', 'frame', 'frameset', 'html', 'iframe', 'input', 'marquee', 'meta', 'noscript', 'object', 'param', 'script', 'style');
 
+	/**
+	 * @var array Stores the default tags to be stripped by rename_attributes().
```
jtojnar, line 654: If I am reading it correctly, the change does not seem to be about tags so the comment and variable names should be updated.

jtojnar, line 654: Maybe use `string[]` as a more detailed type annotation.

library/SimplePie.php
```diff
@@ -1218,6 +1225,15 @@ public function strip_htmltags($tags = '', $encode = null)
 		}
 	}
 
+	public function rename_attributes($tags = '', $encode = null)
```
jtojnar, line 1228: Would be nice to have a doc comment.

library/SimplePie.php
```diff
@@ -1218,6 +1225,15 @@ public function strip_htmltags($tags = '', $encode = null)
...
+	{
+		if ($tags === '')
```
jtojnar, line 1230: Why not `null`? The param could then have `string[]|null` type.

library/SimplePie/Sanitize.php
```diff
@@ -145,6 +146,25 @@ public function pass_file_data($file_class = 'SimplePie_File', $timeout = 10, $u
 		}
 	}
 
+	public function rename_attributes($tags = array())
```
jtojnar, line 149: Maybe we could prefix the method name with `set` to make it clear it is just a setter. I know the old setters do not follow this convention but we could start switching to more modern names (and deprecate the old ones after #711).

library/SimplePie/Sanitize.php
```diff
@@ -145,6 +146,25 @@ public function pass_file_data($file_class = 'SimplePie_File', $timeout = 10, $u
...
+	{
+		if ($tags)
+		{
+			if (is_array($tags))
+			{
+				$this->rename_attributes = $tags;
+			}
+			else
+			{
+				$this->rename_attributes = explode(',', $tags);
```
jtojnar, line 159: Is there any need to support comma separated strings, rather than just arrays of strings? Apps can always do the explosion themselves.

library/SimplePie/Sanitize.php
```diff
@@ -145,6 +146,25 @@ public function pass_file_data($file_class = 'SimplePie_File', $timeout = 10, $u
...
+			}
+		}
+		else
+		{
+			$this->rename_attributes = false;
```
jtojnar, line 164: Why not just use `array()` and have the `foreach` be a no-op?

library/SimplePie.php
math-GH: fixed the comment

math-GH: I want to keep it similar to:

https://github.com/simplepie/simplepie/blob/1aec297145a150f627e6ed82b8b084e0c933def8/library/SimplePie.php#L637-L644

library/SimplePie.php
math-GH: I improved the signature. Second parameter is obsolete.

library/SimplePie/Sanitize.php
math-GH: I agree, it would be better. Let's do it in another PR. I want to focus here on the rename method

library/SimplePie/Sanitize.php
math-GH: It is the same as:
https://github.com/simplepie/simplepie/blob/1aec297145a150f627e6ed82b8b084e0c933def8/library/SimplePie/Sanitize.php#L148-L165

library/SimplePie/Sanitize.php
math-GH: It is the same as:
https://github.com/simplepie/simplepie/blob/1aec297145a150f627e6ed82b8b084e0c933def8/library/SimplePie/Sanitize.php#L148-L165

mblaney:
thanks @math-GH looks good

jtojnar:
The issue with the changes is once we release it, we will not be able able to change it (until 2.0) unless we include more compatibility shims. We should strive to make any new API clearer from the get go, not try to make it look like API that was designed ages ago.

PR #719, opened 2022-02-16
fix file iteration

deminy:
This fix is to address the special case where function call `file_get_contents(".")` and `file_get_contents("..")` are made when running script _build/compile.php_ of version `1.3.x` (or from branch `one-dot-three`).

The calls trigger `E_WARNING` level errors, as explained on [php.net](https://www.php.net/file_get_contents).

### DIFF
--- a/build/compile.php
+++ b/build/compile.php
@@ -51,6 +51,10 @@ $compiled .= remove_header($contents) . "\n";
 $files = new RecursiveIteratorIterator(new RecursiveDirectoryIterator(SP_PATH . '/library/SimplePie'));
 foreach($files as $file_path => $info)
 {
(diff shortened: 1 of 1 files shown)


build/compile.php
```diff
@@ -51,6 +51,10 @@ function remove_header($contents)
 $files = new RecursiveIteratorIterator(new RecursiveDirectoryIterator(SP_PATH . '/library/SimplePie'));
```
jtojnar, line 51: You can just pass `FilesystemIterator::SKIP_DOTS` to [`RecursiveDirectoryIterator`](https://www.php.net/manual/en/class.recursiveiteratoriterator.php).

PR #720, opened 2022-02-18
Bugfix in MySQL cache

Art4:
This PR fixes a bug while saving in MySQL cache.

I've noticed this bug by using [PHPStan](https://phpstan.org/) level 0.


mblaney:
thanks @Art4 !

PR #724, opened 2022-02-27
Add .editorconfig

Alkarex:
See https://editorconfig.org/

Helps text editors automatically adopt the conventions (e.g. for white-space) used in this specific project.

And `.gitattributes` for consistency: ensures line endings are not transformed when working on the same folder from different systems, e.g. WSL / Windows


.editorconfig
```diff
@@ -0,0 +1,27 @@
...
+[*.{markdown,md}]
+indent_size = 4
+indent_style = space
+
+[*.sql]
+indent_style = tab
+
+[*.yml]
+indent_size = 2
+indent_style = space
```
skyzyx, line 27: I would recommend a couple of improvements:

This is the top-level definition, so let's mark it as `root`.

```
root = true
```

You can combine these for efficiency:

```
[*.{css,html,js,json,sql}]
indent_style = tab
```

Use both standard file extensions:

```
[*.{yaml,yml}]
indent_size = 2
indent_style = space
```

Markdown has some additional rules that are important:

```
[*.{markdown,md}]
indent_size = 4
indent_style = space
max_line_length = 0
trim_trailing_whitespace = false
```

Merge these in, and I'm happy to approve.

skyzyx (changes requested):

.editorconfig
Alkarex: I will do as you ask, but here is nevertheless some rationale for the current choices:

* `root = true` prevents users from choosing the display size of their tab indentations (the idea by not having this rule being that users can have an even higher `.editorconfig` with cosmetic preferences)
* `[*.{css,html,js,json,sql}]` fine but makes it more difficult to have distinct rules per language
* `[*.{yaml,yml}]` there is no `*.yaml` file in this project
* `max_line_length` does not exist, does it?
* `trim_trailing_whitespace = false` although possible to want trailing whitespace in Markdown, it is not recommended anymore as far as I know, and I do not think SimplePie makes conscious use of it. e.g. https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md009

Alkarex: @skyzyx Requested changes applied https://github.com/simplepie/simplepie/pull/724/commits/878ce44bcf26f2e8323aa378d7bab9c56f6423be but my comments still stand
As well as for the non-official `max_line_length` (off, not 0) https://github.com/simplepie/simplepie/pull/724/commits/aa11074fe992679978c4a7f72743bfc0321e5477

skyzyx (approved):

.editorconfig
Alkarex: I have just removed the `root` constraint as agreed https://github.com/simplepie/simplepie/pull/724/commits/8d6af85d67dac5a1be7aa66c307086e0ffeeecf9

Regarding whitespace for line breaks, a search shows that this is not something used in the current SimplePie Markdown files. There are five instances of single trailing spaces in https://github.com/simplepie/simplepie/blob/master/CHANGELOG.md but those are clearly mistakes. I would vote for the backslash syntax ` \` or explicit `<br />`.

PR #722, opened 2022-02-21
Use PSR-4 namespaced classes

Art4:
Hey 👋 

Small note about myself: English is not my native language, so I apologize for any language mistakes. I am a long time user of SimplePie directly or indirectly (through WordPress or other libraries). I would like to give something back to the community.

I took some time over the weekend to refactor SimplePie to use namespaces. This means that this PR ~~is not yet ready for merging, but relies on merging #711 and still~~ needs some work and a lot of discussion.

Update: This PR is nearly ready for merge!

## task list

- [x] Merge #711 for having pseudo-classes with namespace in the src folder and aliases created in the library folder.
- [x] move the code from the `library` classes to the namespaced classes
- [x] internally only use the new namespaced classes
- [x] inside the `src` folder don't use any code from the `library` folder
- [x] create aliases in the src folder, e.g. `class_alias('SimplePie\Cache', 'SimplePie_Cache');`
- [x] leave only pseudo-classes or deprecated code in library folder
- [x] move all global `SIMPLEPIE_*` constants into the `SimplePie\SimplePie` class
- [x] set deprecate comment for all global `SIMPLEPIE_*` constants
- [x] update `autoloader.php`
- [x] update `demo/` folder
- [x] update `build/compile.php`
- [x] update README.md
- [x] be more specific in the deprecation notices in `library` classes
- [x] update CHANGELOG.md

## Follow up changes

All the changes in this PR did not change any tests, so all tests (except the new tests from #711) run with the old non-namespaced classes. I intentionally did not change the old tests to show that the old code continues to work as before. These tests need to be modified (or better yet, first copied and then modified) to test the namespaced classes directly. **Update:** See #730

## Possible roadmap

Since I couldn't find any information about the future development of SimplePie or a roadmap, I've sketched out a possible roadmap of how these changes could be released incrementally. If I missed an existing roadmap, I would be happy to be pointed to the right place.

### 1. Why should #711 be released in a minor release?

Following [semantic versioning](https://semver.org/) we have three options:
- The new namespaced classes are keeping the backward compatibility. A new MAJOR version (e.g. v2.0.0) would be need if we make incompatible API changes. ❌
- The new namespaced classes aren't a bugfix so a PATCH version (e.g. v1.5.9) is also not recommended. ❌
- The new namespaced classes are a new feature, because they allow the usage of PSR-4. Adding functionality in a backwards compatible manner should be done in a MINOR version (e.g. v1.6.0) ✅

[...]

PR #725, opened 2022-03-06
Upload compiled file as asset + tests

### LINKED ISSUE #544: Compiled form / build script absent from newer releases
Simplepie 1.3.x used to provide the "compiled" monolithic version directly as a download on the website, which was quite convenient.

I saw this in README.markdown in the 1.5 tarball today:

> If you'd like a single monolithic file, you can run `php build/compile.php` to generate `SimplePie.compiled.php`.

...except the build dir doesn't exist in the release tarballs, so I thought it didn't exist at all, until I found it in git master. So I would think either the compiled form should also be available as a download on the website, or the build dir should be present in the released tarballs, or the README.markdown file should explain where to get it exactly.

Art4:
As I continued to work on #722, I noticed that there are no tests for the output of `build/compile.php`. So I created a new workflow that runs all tests also with the code in the single file. This way we make sure that the single file always works as well. This is especially important when switching to namespaced classes.

Additionally I had the idea, that we can offer the `SimplePie.compiled.php` directly as an asset for download at a release. I have created a separate Github workflow for this purpose.

This PR could also potentially fix #544.


Art4:
I've integrated the tests for the compiled file into the existing ci workflow.

Art4:
Any thoughts or questions? Is there anything I could improve? I would like to continue working on #722, but would need a decision if this PR is accepted.

mblaney:
looks good @Art4 thanks and sorry it took me a while to get to it

Art4:
You're welcome. No problem, I'm patient. 😊

PR #727, opened 2022-04-24
Update notes and copyright date

Art4:
In #711 I used the `build/header.txt` to create the new classes and tests by a script. Now I noted that there is an inappropriate note in all new classes:

>      * Please note: This file is automatically generated by a build script. The
>      * full original source is always available from http://simplepie.org/

This PR removes this note from the files. And because I was already at it, I updated the date of the copyright notice in all places.


Art4:
@mblaney Do you have thoughts on this PR?

mblaney:
@Art4 this looks fine and makes sense to fix.

PR #728, opened 2022-04-25
Fix status_code type

Alkarex:
`$status_code` is declared as `int` but was sometimes `null` or `string`, breaking some contracts.
Downstream: https://github.com/FreshRSS/FreshRSS/pull/4301

### DIFF
--- a/library/SimplePie.php
+++ b/library/SimplePie.php
@@ -429,7 +429,7 @@ class SimplePie
 	 * @see SimplePie::status_code()
 	 * @access private
 	 */
-	public $status_code;
+	public $status_code = 0;
 
 	/**
 	 * @var object Instance of SimplePie_Sanitize (or other class)
--- a/library/SimplePie/File.php
+++ b/library/SimplePie/File.php
@@ -59,7 +59,7 @@ class SimplePie_File
 	var $success = true;
 	var $headers = array();
 	var $body;
-	var $status_code;
+	var $status_code = 0;
 	var $redirects = 0;
 	var $error;
 	var $method = SIMPLEPIE_FILE_SOURCE_NONE;
--- a/library/SimplePie/HTTP/Parser.php
+++ b/library/SimplePie/HTTP/Parser.php
@@ -157,7 +157,7 @@ class SimplePie_HTTP_Parser
 		}
 
 		$this->http_version = '';
-		$this->status_code = '';
+		$this->status_code = 0;
 		$this->reason = '';
 		$this->headers = array();
 		$this->body = '';


Art4 (approved):

jtojnar:
Would not having it type `int|null` and fixing the consumers be cleaner?

Alkarex:
> Would not having it type `int|null` and fixing the consumers be cleaner?

@jtojnar That would change the contract, and my personal preference is towards simpler types rather than more complex types, but besides that I do not really mind. In the case of my client, it would add a bit of code to check the returned type and/or cast to int (null to 0) before sending it to the rest of the pipeline that is expecting an HTTP status to be an integer.

jtojnar:
Yeah, it would change the contract, but so does changing _empty string_ to _zero_. Technically, _zero_ would be an okay choice since it is not actually a valid status code but I generally prefer more explicit types for representing exceptional states over magic values. I have not really looked into the API but I would expect any consumer to filter such states out so they would never make it further in the pipeline.

Alkarex:
> I have not really looked into the API but I would expect any consumer to filter such states out so they would never make it further in the pipeline anyway.

That is the point of the contract. SimplePie has been promising an `int`, which simplifies the code downstream. Having to check for non-int removes most of the advantage of the contract.

I do not really mind, but I believe my current suggestion is marginally more efficient.

Art4:
> Thinking about this more, this is actually a BC break so we might need to change the PHPDoc annotation to int|'' instead, and only fix this in an major bump, if SimplePie follows semver.

Strictly speaking, you are right. But the PHPDoc has always indicated that the value must be an integer. But that sometimes `null` or an empty string is set could be called a bug, which will be fixed now.

In the future you would avoid such bugs by defining the property type. But as long as we can't do that, I would work towards the simplified types.

Alkarex:
> You will almost certainly still need to have something like `if ($simplepie->status_code() === 0) { return; }` to avoid passing invalid status code downstream and having the invalid value as a different type makes it harder to forget about it and then have the downstream function crashing the app.

Just for the record, in the case of my client downstream, the only assumption was I should get an `int` as per contract. A zero is perfectly fine and is not crashing anything (why should it?).

jtojnar:
I would expect any function accepting `int` in lieu of HTTP status code to do basic sanity checks on it and throw an exception otherwise. For example, I just checked the code of a random library and it does precisely that:

https://github.com/guzzle/psr7/blob/5c693242bede743c23402bc5b9de62da04a882d7/src/Response.php#L101

Alkarex:
> I would expect any function accepting `int` in lieu of HTTP status code to do basic sanity checks on it and throw an exception for obviously invalid ones. For example, I just checked the code of a random library and it does precisely that:
> 
> https://github.com/guzzle/psr7/blob/5c693242bede743c23402bc5b9de62da04a882d7/src/Response.php#L101

That is precisely going in the same direction than my argumentation, thanks :-)

`$this->assertStatusCodeRange($status);` calls a function that expects an integer (as it should)....

jtojnar:
Right, the typehint of the `$status` argument of the `Response` constructor only accepts `int`, which means you would not be able to pass it `''` or `null`. That would force you to handle invalid state before even constructing the `Response` object, which is IMO the clean code approach. Additionally, it would make static analysis tools like PHPStan be easily able to point out possible runtime crashes. Whereas when using a magic int value, it is much easier to forget to treat that edge case.

mblaney:
I think this is clear enough, if you want to propose a change @jtojnar I'm happy to look at it but will to merge this for now.

jtojnar:
I opened https://github.com/simplepie/simplepie/pull/732

PR #729, opened 2022-04-25
Fix parsing of HTTP Links

Alkarex:
Before, SimplePie was not able to parse an HTTP header such as:

```
Link: <https://pubsubhubbub.appspot.com>; rel="hub", <https://pubsubhubbub.superfeedr.com>; rel=hub, <https://websubhub.com/hub>; rel="hub"
```

The new regex complies with https://datatracker.ietf.org/doc/html/rfc8288

See also https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link

Downstream: https://github.com/FreshRSS/FreshRSS/pull/4283

### DIFF
--- a/library/SimplePie.php
+++ b/library/SimplePie.php
@@ -2666,12 +2666,14 @@ class SimplePie
 		if (isset($this->data['headers']['link']))
 		{
 			$link_headers = $this->data['headers']['link'];
-			if (is_string($link_headers)) {
-				$link_headers = array($link_headers);
+			if (is_array($link_headers)) {
+				$link_headers = implode(',', $link_headers);
 			}
-			$matches = preg_filter('/<([^>]+)>; rel='.preg_quote($rel).'/', '$1', $link_headers);
-			if (!empty($matches)) {
-				return $matches;
+			// https://datatracker.ietf.org/doc/html/rfc8288
+			if (is_string($link_headers) &&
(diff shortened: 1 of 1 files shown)


library/SimplePie.php
```diff
@@ -2666,12 +2666,14 @@ public function get_links($rel = 'alternate')
...
-			if (is_string($link_headers)) {
-				$link_headers = array($link_headers);
+			if (is_array($link_headers)) {
+				$link_headers = implode(',', $link_headers);
 			}
-			$matches = preg_filter('/<([^>]+)>; rel='.preg_quote($rel).'/', '$1', $link_headers);
-			if (!empty($matches)) {
-				return $matches;
+			// https://datatracker.ietf.org/doc/html/rfc8288
+			if (is_string($link_headers) &&
```
jtojnar, line 2673: Can `$this->data['headers']['link']` be anything other than array or string?

Alkarex: I do not believe so. And this PR does not change the assumptions (there was already a case for array and string before).

jtojnar: In that case, is not `is_string($link_headers)` redundant?

Alkarex: Maybe, but it is not something that is currently enforced (e.g. through a type), and I have not checked in details

mblaney:
thanks @Alkarex

PR #730, opened 2022-04-25
Create tests for PSR-4 classes, refactor oldtests

Art4:
The goal of this PR is to copy and revise as many tests as possible to test against the PSR-4 (namespaced) classes. The old tests for PSR-0 (without namespaces) will remain untouched if possible to ensure BC.

Update: After working my way through the `oldtests` folder, I decided to migrate these very old tests directly after all, to minimize future workload.

**This PR relies on merging #722** and still needs some work.

This PR is a subtask of #731.


Art4:
This PR is ready for review, but still relies on merging of #722.

Art4:
I resolved all merge conflicts after merge of #722.

mblaney:
In general I would say the copyright year shouldn't change when it's the only modification to a file, but all good for now.

Art4:
Thank you. I agree with you in principle about the year but it was a follow up change from #727 to keep all headers consistent.

PR #735, opened 2022-05-19
Do not use magic values for state machine state

jtojnar:
Extracted from https://github.com/simplepie/simplepie/pull/734.

### DIFF
--- a/src/HTTP/Parser.php
+++ b/src/HTTP/Parser.php
@@ -87,11 +87,68 @@ class Parser
 	public $body = '';
 
 	/**
+	 * @access private
+	 */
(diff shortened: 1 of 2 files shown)


Art4:
What do you think about marking the constants as `@access private` or `@internal` to communicate that they are for internal use only and are excluded from the backward compatibility promise.

jtojnar:
Good point, ideally we would add constant visibility modifiers but those are [only available in PHP 7.1](https://stackoverflow.com/questions/5339456/php-class-constant-visibility).

Art4 (approved):

PR #738, opened 2022-05-23
Change code style to PSR-12

Art4:
This PR changes the code style in all classes and tests to [PSR-12](https://www.php-fig.org/psr/psr-12/). As discussed with @jtojnar in https://github.com/simplepie/simplepie/pull/734#issuecomment-1132595248 I think that after the huge changes from #722 the timing is ideal.


Art4:
> * We should also add CI check.

Good point 👍 

> * We probably also want [@PHP54Migration](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/ruleSets/PHP54Migration.rst) and [@PHP56Migration:risky](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/ruleSets/PHP56MigrationRisky.rst)

SimplePie does not support PHP 5.4 anymore. And the only rule in `@PHP56Migration:risky` is `pow_to_exponentiation`, but we don't use `pow()` afaik.
 
> * And maybe even [@Symfony](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/ruleSets/Symfony.rst) or [@Symfony:risky](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/ruleSets/SymfonyRisky.rst) to automate as many style decisions as possible. Though then we will probably want to deactivate [obnoxious rules like `yoda_style`](https://github.com/fossar/selfoss/blob/c0b3a0365464f07bdbb49a98e25eb76db9b36921/.php-cs-fixer.php#L14-L21).

I don't have any particular preferences about this. It's enough for me if rules are defined and these can be implemented automatically by running a script.

jtojnar:
> SimplePie does not support PHP 5.4 anymore

Right, but we did not migrate away from it so we still use long array syntax. [PHP54MigrationSet](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/f736dc928cd803c93bcaadaef24b738e2a0548e4/src/RuleSet/Sets/PHP54MigrationSet.php#L27) would fix that.

Art4:
> Right, but we did not ever migrate to the newly introduced syntax so we still use long array syntax. [PHP54MigrationSet](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/f736dc928cd803c93bcaadaef24b738e2a0548e4/src/RuleSet/Sets/PHP54MigrationSet.php#L27) would fix that.

Oh, wow, you are right. I would have guessed that the short array syntax was already defined in PSR-2. 😅

jtojnar:
PSR does not cover many things we take for granted, which is why I suggested the Symfony ruleset.

jtojnar:
> Do you have a tip or example for me on how to make this happen?

I would probably add `php-cs-fixer:3.8` (pinned version so that CI does not break when new version is released) along with `cs2pr` as described [here](https://github.com/shivammathur/setup-php#tools-with-checkstyle-support) as a new job in https://github.com/simplepie/simplepie/blob/master/.github/workflows/ci.yml.

Having it in composer dev dependencies would be nicer but would force us to stay with old version of php-cs-fixer, as the latest version requires PHP 7.4.

> I would like to move these decisions to a separate PR. In this PR, I would like to have a basic set of rules that everyone can agree on.

On the other hand, we are already changing the coding style significantly (replacing tabs with spaces) so this might be a best time for establishing a coding style once and for all. Each time we change it adds another obstacle to `git blame` (which has already been pretty battered by #722). Once this PR is merged we should add the commit hash to [`.git-blame-ignore-revs`](https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/) and hope it works.

mblaney:
nice work I prefer this coding style too

mblaney:
sorry looks like I missed that this was ready to merge? happy with the composer change but not familiar with the symfony requirements and doesn't feel like something we need to add here?

Art4:
> happy with the composer change but not familiar with the symfony requirements and doesn't feel like something we need to add here?

We could remove the "friendsofphp/php-cs-fixer" library from "require-dev", but it allows every contributor to simply run `bin/php-cs-fixer fix` to fix all CS issues. Without this dependency the `.php-cs-fixer.dist.php` also makes no sense, where the CS rules are defined. I would recommend to keep this dependency.

PR #739, opened 2022-06-19
GH Actions: run tests against PHP 8.2

jrfnl:
The first alpha of PHP 8.2 was released nearly two weeks ago, so let's start testing against PHP 8.2


jrfnl:
Note, it looks like there are some tests missing expectations for a PHP notice, but that is unrelated to this PR.

Issue #731, opened 2022-05-06
Possible roadmap


Art4:
In this issue I collect the progress of the refactoring of SimplePie. This serves roadmap me as a central place for rough planning. Please note that the roadmap is only my personal proposal for the further development of SimplePie, **but has not been officially confirmed by the SimplePie team**.

I would love to hear feedback from the SimplePie team and community. Please share your thoughts on what I think of this roadmap.

## Possible roadmap

All changes in v1.x must be backward compatible to adheres Semantic Versioning. Breaking changes are allowed with the release of v2.

### ✅ SimplePie v1.6.0

- [x] **Merge of #711**.
- [x] Let users try out the new experimantal namespaced classes and collect feedback
- [x] This will also allow us to get a better statistic data about used PHP versions on [packagist.org](https://packagist.org/packages/simplepie/simplepie/php-stats)
- [x] discussion about the necessary changes in `build/compile.php` or the need at all. #725

### 🔲 SimplePie v1.7.0

- [ ] **Merge of #722**.
- [ ] Possibly **raising the minimum PHP version** based on download numbers and feedback, especially from WordPress (ping @rmccue)
- [ ] Copy the tests and change them to test the namespaced classes: #730

### 🔲 SimplePie v1.8.0

- [ ] **Add deprecation warnings** in all classes in `library` folder, see https://github.com/simplepie/simplepie/pull/711#issuecomment-1019526993
- [ ] deprecate the usage of PSR-0
- [ ] deprecate the usage of global constants like `SIMPLEPIE_VERSION`, use `SimplePie\SimplePie::SIMPLEPIE_VERSION` instead
- [ ] possibly deprecate more methods in the `src` folder

### 🔲 SimplePie v2.0.0

- [ ] **remove all deprecated code**
- [ ] remove the `library` folder
- [ ] remove tests for deprecated code
- [ ] Migration to a HTTP library #520 
- [ ] Replace the cache classes with a [PSR-6](https://www.php-fig.org/psr/psr-6/) or [PSR-16](https://www.php-fig.org/psr/psr-16/) implementation

Art4:
The changes for PSR-18 support will take to long, but I would like to release the features for PSR-16 soon. I think it would be better to split the planned features for 1.8.0 into two releases.

jtojnar:
[...]

Issue #520, opened 2017-06-03
Migration to a HTTP library


jtojnar:
Roadmap declares the intention to move from custom web fetcher to something more standardized. At selfoss, we are currently [in the process](https://github.com/SSilence/selfoss/pull/936) of doing the same so we would like to help with this for our dependencies, including SimplePie.

Since selfoss targets PHP 5.4, we would need to use older version of Guzzle (5.*), which might cause conflicts for applications using SimplePie and a different version of Guzzle. For this reason I propose depending on a [HTTPlug](http://httplug.io/) interface and letting the application choose the network library.

If you do not have any objections I will start working on it.

mblaney:
no objections @jtojnar, @JanPetterMG do you use Guzzle with SimplePie? anything to add about implementing this?

JanPetterMG:
I'm using Guzzle, version 6, with custom handlers and middlewares. It opens up a whole new world of options and possibilities, so I'm supporting this migration 110%.

I have a big wish, get rid of the [SimplePie_IRI](https://github.com/simplepie/simplepie/blob/master/library/SimplePie/IRI.php) class. Right now, the Travis builds are failing because of unsupported PHPunit version, but 3 of 4 failures for PHP 7 was due to bugs in that class. _(not sure where the last one was)._ Maybe we'll eventually see official PHP 7 support?

Actually, anyone can use whatever HTTP library they want already now, it's basically this easy:
```php
// Fetch the feed
$simplepie = new \SimplePie();
$simplepie->set_raw_data($xml);
// Parse the `content-type` header and set the encoding
// You MUST skip this if no encoding is provided, or you'll get parsing errors...
$simplepie->set_input_encoding($encoding);
$simplepie->init();
```
The above code has some drawbacks _(nothing that can't be handled externally)_
- Not possible to set an array of urls/feeds, and letting Simplepie merge them (automatically).
- No cache solution built in (`If-Modified-Since`, `ETag`, storage on disk/database)
- No feed auto-discovery, when parsing HTML to detect and redirect itself to an available XML feed

skyzyx:
What I would recommend here is to dig in, and replace the `SimplePie_IRI` class all-together with the newer/better [Intl](http://php.net/Intl) extension.

jtojnar:
For now, I am just overriding the `File` class: https://github.com/SSilence/selfoss/pull/1052

Art4:
[...]

PR #743, opened 2022-08-30
Require PHP 7.2 or newer

Art4:
Hi all 👋 

This PR increases the required PHP version to 7.2 or newer. I propose this PR for SimplePie 1.8 (see Roadmap in #731).

Increasing the PHP version to 7.2 will give us some new features:

- Return type declarations
- Strict typing (`declare(strict_types=1);`)
- Anonymous classes
- Nullable types
- Void functions
- Symmetric array destructuring (already used in this PR)
- Class constant visibility (already used in this PR)
- Abstract method overriding
- Parameter type widening

It will allow us to support PSR-18 HTTP client (see #520)

Based on the download numbers from packagist.org (see below) I would say that it is save to increase the PHP version to 7.2. But I would also like to hear feedback from projects like WordPress (ping @rmccue) and FressRSS (ping @Alkarex).

### Downloads for all SimplePie versions

[Over 96% of all downloads are on PHP 7.2 or newer](https://packagist.org/packages/simplepie/simplepie/php-stats#all).

grafik

### Downloads for SimplePie version 1.6

Looking only at SimplePie 1.6 [over 98% of all downloads are on PHP 7.2 or newer](https://packagist.org/packages/simplepie/simplepie/php-stats#1.6).

grafik


Alkarex:
> But I would also like to hear feedback from projects like WordPress (ping @rmccue) and FressRSS (ping @Alkarex).

We are about to release [FreshRSS 1.20.x supporting PHP 7.0](https://github.com/FreshRSS/FreshRSS/discussions/4517) but from FreshRSS 1.21+ we will require PHP 7.2+

We are more or less tracking Debian and Ubuntu LTS (Debian 9 Stretch supported PHP 7.0 until June 2022)
https://github.com/FreshRSS/FreshRSS/discussions/3321#discussioncomment-835704

Art4:
> P.S. What about WordPress? It is [shipping SimplePie](https://developer.wordpress.org/reference/classes/wp_simplepie_file/), right? And still [supporting PHP 5.6.20+](https://wordpress.org/about/requirements/)

Yes, I would very much appreciate feedback from WordPress. Maybe @jrfnl could help us?

Art4:
Thanks for the detailed explanations @jrfnl. 

I can't speak for the SimplePie team, but based on the roadmap (#731) I can imagine the 1.7 branch as the LTS branch for PHP 5.6 support. If WordPress switches to namespaced classes (`src` folder), backports of bugfixes will be much easier.

In version 1.8 we could bump to PHP 7.2+ and add alternatives to the cache and HTTP client implementations. However, I could also imagine just bumping to PHP 7.0+ instead of 7.2+. This would complicate the upgrade process, but it would allow WordPress to move to a newer version once PHP 6.5 support is dropped.

With 1.9 we could add warnings about deprecated code as a helpful upgrade layer, and with version 2.0 we will remove all deprecated code.

Or to summarize all this in a table:

| Version/branch 	| Status                       	| requires   	| First release      	| Latest patch       	|
|----------------	|------------------------------	|------------	|--------------------	|--------------------	|
| 1.3.x          	| ❌ Unsupported (eol)          	| PHP 5.2.0+ 	| 1.3 (2012-07-07)   	| 1.3.3 (2021-12-24) 	|
| 1.4.x          	| ❌ Unsupported (eol)          	| PHP 5.3.0+ 	| 1.4 (2016-04-26)   	| 1.4.3 (2016-11-27) 	|
| 1.5.x          	| ❌ Unsupported (eol)          	| PHP 5.6.0+ 	| 1.5 (2017-04-17)   	| 1.5.8 (2021-12-21) 	|
| 1.6.x          	| ✅ Supported                  	| PHP 5.6.0+ 	| 1.6.0 (2022-04-21) 	| 1.6.0 (2022-04-21) 	|
| 1.7.x          	| 🏗️ WIP (Switch to PSR-4, LTS) 	| PHP 5.6.0+ 	|                    	|                    	|
| 1.8.x          	| 🏗️ WIP (PSR-16/18 support)    	| PHP 7.0+ or 7.2+ |                    	|                    	|
| 1.9.x          	| 💭 Add deprecation warnings   	| PHP 7.0+ or 7.2+ |                    	|                    	|
| 2.0.x          	| 💭 Remove deprecated code     	| PHP 7.0+ or 7.2+ |                    	|                    	|

jtojnar:
I do not think raising minimum PHP version in minor version would be compatible with semver. It should be done in 2.0.0.

Deprecations should be added in minor versions according to point 7 of [semver](https://semver.org/spec/v2.0.0.html).

So maybe 1.8 + 1.9 should just become 2.0. And 2.0 → 3.0. And 2.0 does not need to have a support window, it can just serve as a migration step.

Art4:
> I do not think raising minimum PHP version in minor version would be compatible with semver. It should be done in 2.0.0.
> 
> Deprecations should be added in minor versions according to point 7 of [semver](https://semver.org/spec/v2.0.0.html).

Increasing the minimum PHP version is not a deprecation, but a change in dependencies. Just as requiring another library as a new dependency is not a breaking change either. Before installing the new version of a library, one must always check if one meets the requirements of the new version. If not, one have to keep the old version. Composer does this automatically and will not allow you to install the new version if the new requirements are not met.

This is also [clarified in semver](https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api).

> So maybe 1.8 + 1.9 should just become 2.0. And 2.0 → 3.0. And 2.0 does not need to have a support window, it can just serve as a migration step.

My proposal is that version 1.9 will be this migration step.

skyzyx:
@jtojnar said:

> I do not think raising minimum PHP version in minor version would be compatible with semver. It should be done in 2.0.0.
> 
> Deprecations should be added in minor versions according to point 7 of [semver](https://semver.org/spec/v2.0.0.html).

SemVer is for defining how a software contract changes over time. However, it does not cover what is in-bounds/out-of-bounds for that software contract. The fact is: SimplePie needs a software contract to go along with SemVer versioning to define exactly this.

My opinion is that since PHP versions go EOL annually, it doesn't make sense to have a major version bump every year. SimplePie 1.0 went live in 2006(?) and supported PHP 4.3. We've dropped several versions of PHP since then.

Also, old versions of SimplePie _still work_. If you are still running an EOL PHP runtime, then you can run an old version of SimplePie. The version you have won't suddenly stop working.

If it were up to me, SimplePie would not consider EOL compatibility in new features or refactorings. PHP 7.4 goes EOL this November/December and only 8.x will be left in supported status. I think it's extremely generous that this PR supports back to PHP 7.2.

To agree with @Art4: We _should_ be dropping support for old versions of PHP annually. If WordPress wants to support PHP 5.6, then they can. But it'll be an older version of SimplePie. While I genuinely **do** appreciate reaching out to affected software projects about compatibility so that everyone can work together, I disagree that those projects should _hold SimplePie back_.

If only we had something like Babel.js for PHP…

jtojnar:
> SemVer is for defining how a software contract changes over time. However, it does not cover what is in-bounds/out-of-bounds for that software contract. The fact is: SimplePie needs a software contract to go along with SemVer versioning to define exactly this.

Good point. In an absence of such contract, I was basing this on best practices from the Node ecosystem, where bumping the major version occurs when a support for a Node dependency is dropped. For PHP, we would need to apply this even for our dependencies since PHP unlike NPM does not support multiple versions of the same library in the dependency tree (unless the symbols are renamed).

> If only we had something like [Babel.js for PHP](https://github.com/rectorphp/rector/issues/7460)…

We do :wink: https://getrector.org/ It works pretty well for me for various refactoring tasks (e.g. https://github.com/j0k3r/graby/pull/282, https://github.com/RSS-Bridge/rss-bridge/pull/2910) and there are people using it to [develop in PHP 8.1 and transpile the artifacts downwards](https://getrector.org/blog/how-all-frameworks-can-bump-to-php-81-and-you-can-use-older-php).

mblaney:
I think it is important to consider WordPress a special case, but the current version of SimplePie included with it is stable, which means WordPress users aren't asking for updates (like has happened previously). To me that means it's ok to make our own decisions and I tend to agree with dropping support for the older versions of PHP.

The recent changes to SimplePie have been great and mostly aimed at modernizing the code base. Seems reasonable given the direction taken that older versions of PHP are dropped so that newer features of the language can be used.

mblaney (approved):

Alkarex:
But so far, SimplePie is only maintaining a single branch, right?
What would be the workflow for e.g. WordPress or others interested in stability rather than fancy new language features, to get new features and security fixes? Fork SimplePie and backport relevant changes?

Art4:
> If we look at [the current stats](https://wordpress.org/about/stats/#php_versions), you can see that PHP 5.6 still represents 5.8% of the user base, PHP 7.0 3,1% and PHP 7.1 2.2%, which comes to a sum-total of > 11% of the users.

Today after 30 days PHP 5.6 has 5,5%, PHP 7.0 has 3,0% and PHP 7.1 has 2,1% user base, which comes to a sum-total of 10,6%. Based on these falling numbers, I would guess that the LTS branch will not be needed for very long.

Art4:
@mblaney What do you think about the idea of a new branching strategy? Now that version 1.7.0 is released, it would be the perfect time to do so. The roadmap for that might look like this:

1. Rename branch `master` to `v1.x` (v1.x will be the new default branch)
2. Create new branch `v1.7-lts` from `v1.x` branch.
3. Edit `README.markdown` to explain new branching strategy. I can do this in this PR.
4. Merge this PR into `v1.x` branch
5. Add note in `v1.7-lts` branch about LTS. I will create a PR for this.

Art4:
I moved the the discussion and necessary changes for a new branching strategy to #751.

mblaney:
this looks good to merge now?

Art4:
> this looks good to merge now?

After the merges of the other PRs I will take a look again for more changes.

tests/Unit/Cache/Psr16Test.php
```diff
@@ -109,7 +109,8 @@ public function testSaveReturnsFalseIfDataOrMtimeCouldNotBeWritten()
 
     public function testSaveCatchesCacheExceptionAndReturnsFalse()
     {
-        $e = $this->createMock(CacheException::class);
+        // BC for PHP <=7.4, use $this->createMock(CacheException::class) in PHP 8.0+
```
jtojnar: Does it even make sense mock exceptions when we do not interact with them? Also, I would expect them to be a stupid objects, so mocking should not be necessary?

PR #749, opened 2022-10-09
Use native array_replace_recursive

Alkarex:
Since PHP 5.3+, there is a native equivalent to the SimplePie Misc function to merge arrays: https://php.net/array-replace-recursive
Improvement of https://github.com/simplepie/simplepie/pull/268

### DIFF
--- a/src/Misc.php
+++ b/src/Misc.php
@@ -204,7 +204,15 @@ class Misc
      */
     public static function array_merge_recursive($array1, $array2)
     {
-        return array_replace_recursive($array1, $array2);
+        foreach ($array2 as $key => $value) {
+            if (is_array($value)) {
+                $array1[$key] = Misc::array_merge_recursive($array1[$key], $value);
+            } else {
+                $array1[$key] = $value;
+            }
+        }
+
+        return $array1;
     }
 
     public static function parse_url($url)
--- a/src/Misc.php
+++ b/src/Misc.php
@@ -199,17 +199,12 @@ class Misc
         return $url;
     }
 
+    /**
+     * @deprecated since SimplePie 1.7.1, use PHP native array_replace_recursive() instead.
+     */
     public static function array_merge_recursive($array1, $array2)
     {
-        foreach ($array2 as $key => $value) {
-            if (is_array($value)) {
-                $array1[$key] = Misc::array_merge_recursive($array1[$key], $value);
-            } else {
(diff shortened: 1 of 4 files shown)


jtojnar:
Note that the functions are not equivalent. For example:

```php
<?php

require_once __DIR__ . '/vendor/autoload.php';

use SimplePie\Misc;

var_dump(Misc::array_merge_recursive(['a' => 'b'], ['a' => ['b']]));
var_dump(array_replace_recursive(['a' => 'b'], ['a' => ['b']]));
```

prints

```php
array(1) {
  ["a"]=>
  string(1) "b"
}
array(1) {
  ["a"]=>
  array(1) {
    [0]=>
    string(1) "b"
  }
}
```

Although the behaviour of `array_replace_recursive` makes more sense to me and there probably won’t be arrays in the values we are merging.

Alkarex:
Good catch :-) I do not think this was the intended behaviour of `Misc::array_merge_recursive()` and was probably an implementation bug (cf. https://github.com/simplepie/simplepie/pull/268 ). The native `array_replace_recursive()` indeed handles arrays better.

src/Misc.php
```diff
@@ -199,17 +199,12 @@ public static function fix_protocol($url, $http = 1)
         return $url;
     }
 
+    /**
+     * @deprecated since SimplePie 1.7.1, use PHP native array_replace_recursive() instead.
+     */
     public static function array_merge_recursive($array1, $array2)
     {
-        foreach ($array2 as $key => $value) {
```
Art4, line 204: For BC reasons I would leave the Misc::array_merge_recursive() method untouched. I think that it is enough if we mark it as deprecated and no longer use it internally.

Alkarex: Done

Art4 (approved):

src/Misc.php
```diff
@@ -199,6 +199,9 @@ public static function fix_protocol($url, $http = 1)
         return $url;
     }
 
+    /**
+     * @deprecated since SimplePie 1.7.1, use PHP native array_replace_recursive() instead.
```
jtojnar, line 203: Maybe also mention that the behaviour is slightly different in case someone really heeds this recommendation?

Art4: I don't think that's necessary. The developer must always check if and how one can use an offered alternative.

PR #750, opened 2022-10-12
remove check for E_USER_DEPRECATED definition

Art4:
`E_USER_DEPRECATED` is always defined since PHP 5.3, therefore, the check is no longer needed.

The checks were introduced when SimplePie still supported PHP 5.2.


jtojnar (approved):
Looks good, https://wiki.php.net/rfc/e-user-deprecated-warning.

PR #752, opened 2022-10-14
Use Psr16 cache as Default, deprecate \SimplePie\Cache\Base implementations

### LINKED ISSUE #616: [BUG] Redis options not set - default to expire 0

### LINKED ISSUE #740: The `file_put_contents()` is not atomic and can cause race in file caching
The problematic code here is here:
https://github.com/simplepie/simplepie/blob/a64bbd4d64a2c97d0dfa99c9a99f296ea2b89e4e/src/Cache/File.php#L114

The problem with that code is that if two PHP processes fetch the file at the same time and use `file_put_contents()` to update the cache, the end result can be interleaved data from both processes which obviously results in invalid cache file which then will later fail when loaded with `file_get_contents()` and `unserialize()`d. This race is pretty hard to reproduce because `file_put_contents()` is basically combination of `open()`+`write()`+`close()` and the race happens if second process can execute `open()` before `write()` is called in first process but I think it happened at least once for our service.

I think something like following would be safe:
```
-                       return (bool) file_put_contents($this->name, $data);
+                       $tmpname = $this->name.".".gethostname()."-".getmypid();
+                       $rv = (bool) file_put_contents($tmpname, $data);
+                       if (false === $rv)
+                               return false;
+                       $rv = rename($tmpname, $this->name);
+                       return $rv;
```
This includes `gethostname()` to make the execution still atomic if cache is backed by NFS and two hosts running SimplePie happened to be unlucky enough to use the same PID, too. As far as I know, if the backend is NFS v2 there's no way to get filesystem to reserve atomic unique temporary filename so suitable filenames should be used instead. Even with old NFS backends, `rename()` is atomic meaning either the old version is active or new version is active but there can never be mixed case in between.

Basically the above implementation writes data to temporary filename first where temporary filename includes hostname and PID of the currently running process. This should result in a filename that cannot ever collide with concurrent PHP process and the `rename()` is atomic resulting in one intact cache file to remain in use.

This implementation would leave the stale temporary file if `rename()` fails for any reason. I'm not sure if that's good or bad. One could try to call `unlink()` for the temporary file if `rename()` fails but I'd expect `rename()` fail only in case of read-only filesystem and `unlink()` would then obviously fail, too. However, if you assume permissions of existing cache file has been set incorrectly, then `rename()` might fail because of file permissions – leaving the temporary file around in this case might be beneficial if admin were then to fix the issue manually.

And I'm not sure if `gethostname()` can actually fail to return a string in practice. According to the documentation it can return false. That would catenate as empty string so it should be still okay in the above implementation but that could result in identical temporary filename used on NFS backend.

Art4:
Hi all 👋 

This is a follow-up PR of #742. I propose this PR for SimplePie 1.8, see #731.

In this PR I've changed the internal cache usage to a more PSR-16 style, see the new interface `SimplePie\Cache\DataCache`. This interface could be removed in SimplePie 2.0 and the provided PSR-16 implementation can be used directly. This will bring us more advantages like `\Psr\SimpleCache\CacheInterface::setMultiple()`.

For BC reasons I've created a simple adapter class `SimplePie\Cache\BaseDataCache` for the current `SimplePie\Cache\Base` implementations. This allows us to use the PSR-16 implementation from #742, but also support the old cache implementations. Furthermore I deprecated all cache implementations in favor of the PSR-16 cache.

This will 
- fix #740, 
- fix #616 
- and make obsolete #695.


src/Cache/DataCache.php
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * @param string   $key     The unique key of this item in the cache.
+     * @param mixed    $default Default value to return if the key does not exist.
+     *
+     * @return array|mixed The value of the item from the cache, or $default in case of cache miss.
+     *
+     * @throws InvalidArgumentException
+     *   MUST be thrown if the $key string is not a legal value.
+     */
+    public function getData($key, $default = null);
```
Art4, line 75: I think I should change the methods to camel_case.

jtojnar: On another hand, switching to `camelCase` might give us opportunity to clean up the API. The new methods can use native types (e.g. `DateTimeImmutable` instead of `get_date`’s `int|string`) and once we drop PHP 5 support, native return typehints.

Art4: Good point. Using the switch from snake_case to camelCase to introduce return types is a very good idea. But because this will require the merge of #743, the decision to switch to camelCase in v2 and complete new interfaces I would recommend to use snake_case for now. I put this idea to the roadmap #731.

jtojnar: For `final` classes (or, I guess, `internal` ones too), we should not need to worry about [variance](https://www.php.net/manual/en/language.oop5.variance.php) so it should not matter if we add typehints later. That should allows to use camel case here now.

src/Cache/DataCache.php
```diff
@@ -0,0 +1,114 @@
...
+use InvalidArgumentException;
+
+/**
+ * Simplified PSR-16 Cache client for caching data arrays
+ *
+ * The methods names must be different, but should be compatible to the
+ * methods of \Psr\SimpleCache\CacheInterface.
+ *
+ * @package SimplePie
+ * @subpackage Caching
```
jtojnar, line 55: Probably should be internal too?

src/Cache/DataCache.php
Art4: Yes, you are right. But the decision to switch to camelCase has to be discussed and made first in another issue. Then I can still change the PR.

src/Cache/DataCache.php
Art4: done

jtojnar:
Might be nice to rebase this onto master so it can be reviewed commit by commit more easily.

Art4:
Sorry, I tried to rebase but I failed. I don't have much experience with rebase (maybe because I prefer merge 😊) and the conflicts are to complicated for me.

CHANGELOG.md
```diff
@@ -11,6 +11,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - New method `SimplePie\SimplePie::set_cache()` for providing a PSR-16 cache implementation
 
+### Deprecated
+
+- The method `SimplePie\SimplePie::set_cache_location()` is deprecated, use `SimplePie\SimplePie::set_cache()` instead
+- The method `SimplePie\SimplePie::force_cache_fallback()` is deprecated, expired cache will not be used anymore
+- The class `SimplePie\Cache` is deprecated, use implementation of `SimplePie\SimplePie::set_cache()` instead
```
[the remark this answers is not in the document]
Art4: I thought while using `SimplePie\Cache` will not only allow you to get a cache implementation but also to register a cache implementation the reference to `SimplePie\SimplePie::set_cache()` makes that clearer.

src/Cache/BaseDataCache.php
```diff
@@ -0,0 +1,157 @@
+<?php
+/**
+ * SimplePie
```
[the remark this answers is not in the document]
Art4: Never heard of reuse.software before, but it sounds good.

src/Cache/DataCache.php
```diff
@@ -0,0 +1,115 @@
...
+ * @link http://simplepie.org/ SimplePie
+ * @license http://www.opensource.org/licenses/bsd-license.php BSD License
+ */
+
+namespace SimplePie\Cache;
+
+use InvalidArgumentException;
+
+/**
+ * Simplified PSR-16 Cache client for caching data arrays
```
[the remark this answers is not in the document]
Art4: It uses only a subset of PSR-16 (only `get()`, `set()` and `delete()`, but not `has()`, `getMultiple()`, `setMultiple()` or `deleteMultiple()`),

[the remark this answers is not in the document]
jtojnar: Might be nice to mention it explicitly in the comment.

src/Cache/BaseDataCache.php
```diff
@@ -0,0 +1,157 @@
...
+     */
+    public function get_data($key, $default = null)
+    {
+        $data = $this->cache->load();
+
+        if (! is_array($data)) {
+            return $default;
+        }
+
+        // ingore data if internal cache expiration time is not set
```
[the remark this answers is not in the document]
Art4: fixed in #761

src/Cache/DataCache.php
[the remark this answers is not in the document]
Art4: fixed in #761

PR #758, opened 2022-10-24
Add HTTP layer to encapsulate SimplePie\File

Art4:
This PR will add a new layer for all HTTP requests to encapsulate the usage of the `SimplePie\File`.

I propose this PR for SimplePie 1.8.0, see #731. This also refs #520.

I created the new `SimplePie\HTTP\Client` and `SimplePie\HTTP\Response` interfaces inspired by [PSR-18](https://www.php-fig.org/psr/psr-18/). Next I created the adapter `SimplePie\HTTP\FileClient` to allow HTTP requests via `SimplePie\File` class. In a followup PR a new adapter will be possible to use a PSR-18 implementation directly.

Encapsulate the usage of the `SimplePie\File` while keeping BC seems to be a complicated task. Therefor I've created this PR in an early draft stage. I invite everyone to give me early feedback.


Art4:
This PR will become to large and complex for reviews. Therefor I will try to split the PR into multiple smaller PRs. This branch will serve me for further experiments on the poc, but I close the PR for now.

PR #759, opened 2022-10-30
Type declarations in cache

Art4:
This PR will add parameter and return type declarations to the new internal `DataCache` interface from #752. It will also fix the setting of two `$ttl`.

### DIFF
--- a/src/Cache/BaseDataCache.php
+++ b/src/Cache/BaseDataCache.php
@@ -80,7 +80,7 @@ final class BaseDataCache implements DataCache
      * @throws InvalidArgumentException
      *   MUST be thrown if the $key string is not a legal value.
      */
-    public function get_data($key, $default = null)
+    public function get_data(string $key, $default = null)
     {
         $data = $this->cache->load();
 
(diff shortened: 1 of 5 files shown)


src/Cache/BaseDataCache.php
```diff
@@ -80,7 +80,7 @@ public function __construct(Base $cache)
      * @throws InvalidArgumentException
```
jtojnar, line 80: What do you mean by “setting of two `$ttl`”?

src/Cache/BaseDataCache.php
```diff
@@ -123,7 +123,7 @@ public function get_data($key, $default = null)
...
      *   MUST be thrown if the $key string is not a legal value.
      */
-    public function set_data($key, array $value, $ttl = null)
+    public function set_data(string $key, array $value, ?int $ttl = null): bool
     {
         if (! is_int($ttl)) {
```
jtojnar, line 128: This will allow us to simplify:

```suggestion
        if ($ttl === null) {
```

src/Cache/BaseDataCache.php
Art4: I had specified the `$ttl` in SimplePie.php as `$this->cache_duration + time()`. This means a period of over 52 years that the cache would be valid.

jtojnar (approved):

PR #762, opened 2022-11-03
Replace cache_name_function with NameFilter class

Art4:
It is possible to change the name of a cached value by providing a callable string via `SimplePie\SimplePie::set_cache_name_function()`. The default function is `md5()`. This gives us two problems: 

1. The provided callable is not guarantied to return a string. This could lead to errors.
2. A callable could also be an array. Something like this is not possible:

```php
$simplepie->set_cache_name_function(['Classname', 'static_method']);
```

This PR adds a new interface `SimplePie\Cache\NameFilter` that will be used internally. Implementations can be set via the new method `SimplePie\SimplePie::set_cache_namefilter()`. For BC reasons there is also a new class `SimplePie\Cache\CallableNameFilter`, that handles all kinds of callables (not just strings).

```php
$simplepie->set_cache_namefilter(new \SimplePie\Cache\CallableNameFilter(['Classname', 'static_method']);
```

This PR also deprecates the `SimplePie\SimplePie::set_cache_name_function()` method.

### DIFF
--- a/src/Cache/CallableNameFilter.php
+++ b/src/Cache/CallableNameFilter.php
@@ -1,0 +1,75 @@
+<?php
+/**
+ * SimplePie
+ *
+ * A PHP-Based RSS and Atom Feed Framework.
+ * Takes the hard work out of managing a complete RSS/Atom solution.
+ *
+ * Copyright (c) 2004-2022, Ryan Parman, Sam Sneddon, Ryan McCue, and contributors
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are
+ * permitted provided that the following conditions are met:
+ *
+ * 	* Redistributions of source code must retain the above copyright notice, this list of
+ * 	  conditions and the following disclaimer.
+ *
+ * 	* Redistributions in binary form must reproduce the above copyright notice, this list
+ * 	  of conditions and the following disclaimer in the documentation and/or other materials
+ * 	  provided with the distribution.
+ *
+ * 	* Neither the name of the SimplePie Team nor the names of its contributors may be used
+ * 	  to endorse or promote products derived from this software without specific prior
+ * 	  written permission.
+ *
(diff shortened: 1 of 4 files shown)


jtojnar (commented):
I am not sure the added boilerplate is worth it for an obscure function like this.

Hopefully, PHP will eventually add support for generics so we will be able to annotate the argument with `callable(string): string` natively. For now, the combination of static analysis and runtime crashes (which will occur earlier once we enable [strict type checking](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict)).

Actually, I am not even sure if SimplePie needs to make this customizable.

Art4:
I think I get your point. Unfortunately, we cannot know, if and how this function is used, so at least we could add stricter rules.

On the other hand I can see a legit use case for this function. Especially in view of the fact that we give the user the possibility to use his own PSR-16 cache. PSR-16 provides [some rules for the cache key](https://www.php-fig.org/psr/psr-16/#12-definitions) (only specific characters, not more than 64 chars, and so on), which are fulfilled with `md5()`. But a provided PSR-16 cache is allowed to widening this rules.

Examples:

**Use longer cache key**

```php
$simplepie->set_cache_namefilter(new class () implements \SimplePie\Cache\NameFilter {
    public function filter(string $name): string {
        return hash('sha3-512', $name);
    }
});
```

**Add prefix to cache key**

```php
$simplepie->set_cache_namefilter(new class () implements \SimplePie\Cache\NameFilter {
    public function filter(string $name): string {
        return 'simplepie_' . hash('md5', $name);
    }
});
```

**Use raw key and handle everything in the PSR-16 cache**

```php
$simplepie->set_cache_namefilter(new class () implements \SimplePie\Cache\NameFilter {
    public function filter(string $name): string {
        return $name;
    }
});
```

In my opinion this function offers the best flexibility for all use cases.

Art4:
Ping @mblaney

mblaney:
looks good but there's a conflict now from the other merges

Art4:
Thank you. I fixed the conflict.

PR #763, opened 2022-11-04
Declare strict_types=1 in every file

Art4:
By default, PHP will coerce values of the wrong type into the expected scalar type declaration if possible. It is possible to enable [strict mode](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict), so only a value corresponding exactly to the type declaration will be accepted, otherwise a `TypeError` will be thrown.

This PR enables strict mode in all files. In most of the classes, everything works as expected but in a few classes I had to cast some values explicitly. I could even find and fix some bugs:

- Fixed a [bug in the `Misc` class](https://github.com/simplepie/simplepie/pull/763/files#diff-3dd17bdf04ae7a362e844d2f16376a40b4cfbbe707f512a02c1cd8831419b7d0L1802)
- Fixed `SimplePie::get_image_height()` and `get_image_width()` returning `int|null` instead of `int|float|null`, introduced in https://github.com/simplepie/simplepie/commit/33a8f76cdf6eb55131d8d9616034661c04882b9f

I've also checked and fixed most of the errors for PHPStan Level 1.

I will keep this PR as draft until the other open PRs are merged so I can later add the strict types in the new classes and interfaces created by theses PRs.

### DIFF
--- a/src/HTTP/Parser.php
+++ b/src/HTTP/Parser.php
@@ -1,4 +1,6 @@
 <?php
+
+declare(strict_types=1);
 /**
  * SimplePie
  *
--- a/src/HTTP/Parser.php
+++ b/src/HTTP/Parser.php
@@ -472,7 +472,7 @@ class Parser
             $encoded = substr($encoded, $chunk_length + $length + 2);
 
             // BC for PHP < 8.0: substr() can return bool instead of string
-            $encoded === false ? '' : $encoded;
+            $encoded = ($encoded === false) ? '' : $encoded;
 
             if (trim($encoded) === '0' || empty($encoded)) {
                 $this->state = self::STATE_EMIT;
--- a/src/HTTP/Parser.php
+++ b/src/HTTP/Parser.php
@@ -468,8 +468,11 @@ class Parser
             }
 
             $chunk_length = strlen($matches[0]);
-            $decoded .= $part = substr($encoded, $chunk_length, $length);
+            $decoded .= substr($encoded, $chunk_length, $length);
             $encoded = substr($encoded, $chunk_length + $length + 2);
+
+            // BC for PHP < 8.0: substr() can return bool instead of string
+            $encoded = strval($encoded);
 
             if (trim($encoded) === '0' || empty($encoded)) {
                 $this->state = self::STATE_EMIT;
--- a/src/HTTP/Parser.php
+++ b/src/HTTP/Parser.php
@@ -472,7 +472,7 @@ class Parser
             $encoded = substr($encoded, $chunk_length + $length + 2);
 
             // BC for PHP < 8.0: substr() can return bool instead of string
-            $encoded = strval($encoded);
+            $encoded === false ? '' : $encoded;
 
             if (trim($encoded) === '0' || empty($encoded)) {
                 $this->state = self::STATE_EMIT;
--- a/src/IRI.php
+++ b/src/IRI.php
@@ -378,7 +378,7 @@ class IRI
                 $output = substr_replace($output, '', intval(strrpos($output, '/')));
             } elseif ($input === '/..') {
                 $input = '/';
-                $output = substr_replace($output, '', strrpos($output, '/'));
+                $output = substr_replace($output, '', intval(strrpos($output, '/')));
             }
             // D: if the input buffer consists only of "." or "..", then remove that from the input buffer; otherwise,
             elseif ($input === '.' || $input === '..') {
--- a/src/IRI.php
+++ b/src/IRI.php
@@ -545,7 +545,7 @@ class IRI
         $start = 0;
         $character = 0;
         $length = 0;
-        $valid = true; // By default we are valid
+        $valid = true;
 
         // Loop over each and every byte, and set $value to its value
         for ($i = 1, $len = count($bytes); $i < $len; $i++) {
@@ -555,6 +555,9 @@ class IRI
             if (!$remaining) {
                 // Start position
                 $start = $i;
+
+                // By default we are valid
+                $valid = true;
 
                 // One byte sequence:
                 if ($value <= 0x7F) {
--- a/src/IRI.php
+++ b/src/IRI.php
@@ -1,4 +1,6 @@
 <?php
+
+declare(strict_types=1);
 /**
  * SimplePie
  *
@@ -373,7 +375,7 @@ class IRI
             // C: if the input buffer begins with a prefix of "/../" or "/..", where ".." is a complete path segment, then replace that prefix with "/" in the input buffer and remove the last segment and its preceding "/" (if any) from the output buffer; otherwise,
             elseif (strpos($input, '/../') === 0) {
                 $input = substr($input, 3);
-                $output = substr_replace($output, '', strrpos($output, '/'));
+                $output = substr_replace($output, '', intval(strrpos($output, '/')));
             } elseif ($input === '/..') {
                 $input = '/';
                 $output = substr_replace($output, '', strrpos($output, '/'));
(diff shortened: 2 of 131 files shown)


src/SimplePie.php
```diff
@@ -2755,7 +2757,7 @@ public function get_image_link()
     public function get_image_width()
     {
         if ($return = $this->get_image_tags(self::NAMESPACE_RSS_20, 'width')) {
-            return round($return[0]['data']);
+            return round(intval($return[0]['data']));
```
jtojnar, line 2760: No point of rounding an int.

src/Parse/Date.php
```diff
@@ -726,7 +728,7 @@ public function date_w3cdtf($date)
             // Convert the number of seconds to an integer, taking decimals into account
             $second = round((int)$match[6] + (int)$match[7] / (10 ** strlen($match[7])));
 
-            return gmmktime($match[4], $match[5], $second, $match[2], $match[3], $match[1]) - $timezone;
+            return gmmktime(intval($match[4]), intval($match[5]), intval($second), intval($match[2]), intval($match[3]), intval($match[1])) - $timezone;
```
jtojnar, line 731: Ideally, we would convert it into a proper type as soon as possible (e.g. https://github.com/simplepie/simplepie/pull/765) but this looks good for now.

src/HTTP/Parser.php
```diff
@@ -468,9 +468,12 @@ protected function chunked()
             }
 
             $chunk_length = strlen($matches[0]);
-            $decoded .= $part = substr($encoded, $chunk_length, $length);
+            $decoded .= substr($encoded, $chunk_length, $length);
             $encoded = substr($encoded, $chunk_length + $length + 2);
 
+            // BC for PHP < 8.0: substr() can return bool instead of string
+            $encoded = strval($encoded);
```
jtojnar, line 475: Again, I would use something like `$encoded === false ? '' : $encoded`.

src/IRI.php
```diff
@@ -539,19 +540,22 @@ protected function remove_iunreserved_percent_encoded($match)
...
+        $start = 0;
+        $character = 0;
+        $length = 0;
 
         // Loop over each and every byte, and set $value to its value
         for ($i = 1, $len = count($bytes); $i < $len; $i++) {
             $value = hexdec($bytes[$i]);
 
+            // By default we are valid
+            $valid = true;
```
jtojnar, line 552: While I do not think this changes the semantics, it might be clearer to only reset this when an initial byte is read.

src/IRI.php
```diff
@@ -539,19 +540,22 @@ protected function remove_iunreserved_percent_encoded($match)
         // at the first byte!).
         $string = '';
         $remaining = 0;
+        $start = 0;
+        $character = 0;
+        $length = 0;
```
jtojnar, line 545: Instead I would add extra `$valid = true` here to satisfy PHPStan. And add a comment that these variables will be initialized in the loop but PHPStan is not able to detect it ([currently](https://github.com/phpstan/phpstan/issues/8192)).

src/IRI.php
```diff
@@ -373,7 +375,7 @@ protected function remove_dot_segments($input)
             // C: if the input buffer begins with a prefix of "/../" or "/..", where ".." is a complete path segment, then replace that prefix with "/" in the input buffer and remove the last segment and its preceding "/" (if any) from the output buffer; otherwise,
             elseif (strpos($input, '/../') === 0) {
                 $input = substr($input, 3);
-                $output = substr_replace($output, '', strrpos($output, '/'));
+                $output = substr_replace($output, '', intval(strrpos($output, '/')));
```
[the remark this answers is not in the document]
Art4: My motivation to improve this soon to be deprecated class is not very high. 😅

src/IRI.php
Art4: done.

src/IRI.php
Art4: done.

src/IRI.php
```diff
@@ -817,7 +819,7 @@ public function set_authority($authority, $clear_cache = false)
         } else {
             $iuserinfo = null;
         }
-        if (($port_start = strpos($remaining, ':', strpos($remaining, ']'))) !== false) {
+        if (($port_start = strpos($remaining, ':', intval(strpos($remaining, ']')))) !== false) {
```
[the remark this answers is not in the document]
Art4: May I ask for an example?

src/IRI.php
```diff
@@ -547,9 +556,6 @@ protected function remove_iunreserved_percent_encoded($match)
                 // Start position
                 $start = $i;
 
-                // By default we are valid
-                $valid = true;
```
jtojnar, line 551: This is still necessary, otherwise invalid parse will make rest of the string invalid.

src/SimplePie.php
```diff
@@ -2757,7 +2757,7 @@ public function get_image_link()
     public function get_image_width()
     {
         if ($return = $this->get_image_tags(self::NAMESPACE_RSS_20, 'width')) {
-            return round(intval($return[0]['data']));
+            return intval($return[0]['data']);
```
jtojnar, line 2760: I was thinking moving the `round` inside `intval` so that the value is int but properly rounded.

src/IRI.php
Art4: Sorry, you are right. This should be fixed now.

src/SimplePie.php
jtojnar: Yeah, I would assume that is why `round` was used in the first place.

jtojnar: But maybe it is not worth it.

Art4: Before we can use `round()` we would convert it to float first. It would look like this: 
```php
return intval(round(floatval($return[0]['data'])));
```

see also: https://3v4l.org/sHMLb

I don't think (and hope) that anybody would provide a size as "half pixels". 😄 Or at least would don't have to support this. According to [RSS 2.0 spec](https://www.rssboard.org/rss-specification#ltimagegtSubelementOfLtchannelgt) the pixels must be "numbers".

jtojnar:
@mblaney Could you please create `one-dot-eight` branch from `1.8.0` so that we could revert this PR there for `1.8.1` to fix #810?

Art4:
Why should we revert the complete PR? A better option would be to remove `declare(strict_types=1);` from a specific file, instead from all files.

jtojnar:
Because it is impossible to tell which files are actually broken without a complete test coverage or static analysis. And since most of our functions currently only have argument types specified using a PHPDoc, incorrectly typed value can cause an error arbitrarily deep, when they are passed to native PHP functions.

So far we have seen four or five different instances of this, IIRC, and I doubt those are all of the issues.

Art4:
> Because it is impossible to tell which files are actually broken without a complete test coverage or static analysis. And since most of our functions currently only have argument types specified using a PHPDoc, incorrectly typed value can cause an error arbitrarily deep, when they are passed to native PHP functions.
> 
> So far we have seen four or five different instances of this, IIRC, and I doubt those are all of the issues.

I would like to see bug fixes releases more often where the issues are fixed instead of removing the declare statement.. But I'm afraid that the chances are not very high that this will happen. Maybe we could speed up the release for 1.9.0 instead?

jtojnar:
> I would like to see bug fixes releases more often where the issues are fixed instead of removing the declare statement.. But I'm afraid that the chances are not very high that this will happen.

Same here.

> Maybe we could speed up the release for 1.9.0 instead?

Well, I see the following options:

- Remove `strict_types` everywhere until we have high certainty that the issues do not occur.
  - This is the easiest solution and could be even done in 1.8 branch.
- Remove `strict_types` just from files where the issue has been reported.
  - Potentially will need to be repeated as new reports come.
- Look for all files where the problematic calls are and remove `strict_types`.
  - How? Maybe functions passing values of untyped arguments to native PHP functions? Sounds like a lot of work. Probably best left for static analysis.
- Add tests for all possible argument types of all functions.
  - Even 100 % line coverage is not guarantee of lack of errors. But might still give us some degree of assurance.
  - Still lot of work.
- Raise PHPStan level (IIRC, 8 or 9 should catch these kind of issues)
  - This will also take time but I think this is best option for 1.9.0.

PR #761, opened 2022-11-03
Release 1.8.0

Art4:
In this PR we can prepare everything for the release of SimplePie 1.8.0

Check list:

- [x] bump version in constant `SimplePie\SimplePie::VERSION`.
- [x] Add entries for all changes in `CHANGLOG.md`.
- [x] Add missing authors and issues in `CHANGLOG.md` if necessary.
- [x] Fix code style by running `bin/php-cs-fixer fix`.

PR #764, opened 2022-11-07
Remove broken BC code in Registry

Art4:
While working on the `Registry` class I noted this BC code that actually never worked. It was introduced in https://github.com/simplepie/simplepie/commit/f6e63e6ba8e4a198283bb2323aa498fd0d6a216e (version 1.3).

The code was intended to fix a BC break in the `Locator::__construct()` method. But because the switch looks for `locator` instead of (uppercase) `Locator` there was no chance to register a BC class that matches this requirement. Please note also the wrong lower case in `$this->get_class('file')` and `$this->get_class('content_type_sniffer')` that would always return `null` instead of a class name. Also I could not find any complaints about a BC break for the changed parameter in the `Locator::__construct()` method. 

Therefor I propose to remove this BC code, that actually never worked.

### DIFF
--- a/src/Registry.php
+++ b/src/Registry.php
@@ -160,17 +160,6 @@ class Registry
     {
         $class = $this->get_class($type);
 
-        if (in_array($class, $this->legacy)) {
-            switch ($type) {
-                case 'locator':
-                    // Legacy: file, timeout, useragent, file_class, max_checked_feeds, content_type_sniffer_class
-                    // Specified: file, timeout, useragent, max_checked_feeds
-                    $replacement = [$this->get_class('file'), $parameters[3], $this->get_class('content_type_sniffer')];
-                    array_splice($parameters, 3, 1, $replacement);
-                    break;
-            }
-        }
-
         if (!method_exists($class, '__construct')) {
             $instance = new $class();
         } else {


jtojnar (approved):
Looks good – the only way this could work if someone used the Registry newly introduced in https://github.com/simplepie/simplepie/commit/f6e63e6ba8e4a198283bb2323aa498fd0d6a216e with the legacy class and lowercase name `$simplepie->registry->register('locator', $class, true)`. But as you noticed that is not possible, since the lowercase `$type` argument would not be a key in the `Registry::$defaults` property so registration would fail on not finding a parent class to check subclass relationship for.

jtojnar:
Now, if only we could drop the other legacy alias:

https://github.com/Art4/simplepie/commit/5f107f3bd02918e296a856b322da49235e6bcbb1#diff-c09efaf19f27f39a1f6829e8ded114062e329cb15f8ce1d65f1f0a6c6bb159bd

Unfortunately, that one can be relied on externally so we would need to deprecate it first. While at it, we should probably also deprecate `Simplepie::set_*_class()` methods, since that is the place that legacy classes are registered.

Art4:
> Now, if only we could drop the other legacy alias:
> 
> [5f107f3#diff-c09efaf19f27f39a1f6829e8ded114062e329cb15f8ce1d65f1f0a6c6bb159bd](https://github.com/simplepie/simplepie/commit/5f107f3bd02918e296a856b322da49235e6bcbb1#diff-c09efaf19f27f39a1f6829e8ded114062e329cb15f8ce1d65f1f0a6c6bb159bd)

I also thought about removing that, but since the whole `Cache` class is already deprecated, it won't bother me.

> Unfortunately, that one can be relied on externally so we would need to deprecate it first. While at it, we should probably also deprecate `Simplepie::set_*_class()` methods, since that is the place that legacy classes are registered.

The `Simplepie::set_*_class()` methods are already deprecated since SimplePie 1.3, see the phpdoc template in https://github.com/simplepie/simplepie/blob/1.3/library/SimplePie.php#L887-L894
I made that clearer in #761.

mblaney:
thanks @Art4 good catch!

PR #765, opened 2022-11-13
Parse\Date: Clean up regex structure

jtojnar:
Looks like there are no tests 🤷‍♀

### DIFF
--- a/src/Parse/Date.php
+++ b/src/Parse/Date.php
@@ -679,54 +679,66 @@ class Date
      */
     public function date_w3cdtf($date)
     {
-        static $pcre;
-        if (!$pcre) {
-            $year = '([0-9]{4})';
-            $month = $day = $hour = $minute = $second = '([0-9]{2})';
-            $decimal = '([0-9]*)';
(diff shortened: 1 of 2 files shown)


src/Parse/Date.php
```diff
@@ -679,54 +679,66 @@ public function add_callback($callback)
...
-            } else {
-                $timezone = 0;
-            }
+            $timezone = $tz_hour * 3600;
+            $timezone += $tz_minute * 60;
+            $timezone *= $tz_sign;
 
             // Convert the number of seconds to an integer, taking decimals into account
-            $second = round((int)$match[6] + (int)$match[7] / (10 ** strlen($match[7])));
+            $second = round($second + $second_fraction);
```
Art4, line 739: We should cast the returned `float` from `round()` into `int`.

```suggestion
            $second = (int) round($second + $second_fraction);
```

Art4 (changes requested):
I like the well formated `$pcre` with named subpatterns very much. 😃 And special 👍 for adding the tests.

src/Parse/Date.php
jtojnar: Fixed, thanks.

Art4 (approved):

PR #766, opened 2022-11-17
Registry: Allow using class-strings instead of magic strings

Art4:
As discussed with @jtojnar I finalized #737 for using class-strings instead of strings in `Registry` class. #737 can be closed.

It also create legacy replacements to provide BC. In future it should also possible to use an interface as `$type` and provide an implementing class as `$class`.

### DIFF
--- a/src/Registry.php
+++ b/src/Registry.php
@@ -99,7 +99,15 @@ class Registry
      * @see register()
      * @var array<string, class-string>
      */
-    protected $legacy = [
+    protected $legacy = [];
+
+    /**
+     * Legacy types
+     *
+     * @see register()
+     * @var array<string, class-string>
+     */
+    private $legacyTypes = [
         'Cache' => Cache::class,
         'Locator' => Locator::class,
         'Parser' => Parser::class,
@@ -140,7 +148,11 @@ class Registry
      */
     public function register($type, $class, $legacy = false)
     {
-        if (! array_key_exists($type, $this->default) && ! array_key_exists($type, $this->legacy)) {
+        if (array_key_exists($type, $this->legacyTypes)) {
+            $type = $this->legacyTypes[$type];
+        }
+
+        if (! array_key_exists($type, $this->default)) {
             return false;
         }
 
@@ -149,7 +161,7 @@ class Registry
         }
 
         /** @var string */
-        $base_class = $this->default[$type] ?? $this->legacy[$type];
+        $base_class = $this->default[$type];
 
         if (!@is_subclass_of($class, $base_class)) {
             return false;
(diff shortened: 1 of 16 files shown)


src/Registry.php
```diff
@@ -156,21 +178,13 @@ public function get_class($type)
...
-        }
-
-        // Try again with the legacy type resolved.
-        $type = $this->legacyClassNames[$type];
-        if (!empty($this->classes[$type])) {
-            return $this->classes[$type];
-        }
-        if (!empty($this->default[$type])) {
-            return $this->default[$type];
+        if (!empty($this->legacy[$type])) {
```
jtojnar: This will no longer return properly registered class:

```php
class CustomSniffer extends SimplePie\Content\Type\Sniffer {}

$registry->register(SimplePie\Content\Type\Sniffer::class, CustomSniffer::class);
var_dump($registry->get_class(`Content_Type_Sniffer`) instanceof CustomSniffer); // returns false
```

tests/Unit/SimplePieTest.php
```diff
@@ -264,11 +269,6 @@ public function testLegacyCallOfSetCacheClass()
             $this->expectError();
         }
 
-        $feed = new SimplePie();
```
jtojnar, line 267: Why this unrelated change?

src/Registry.php
```diff
@@ -53,41 +57,33 @@
...
+     * Default class mapping
      *
+     * Overriding classes *must* subclass these.
      *
-     * @var array<string, class-string>
+     * @var array<class-string, class-string>
      */
-    protected $default = [];
+    protected $default = [
+        Cache::class => Cache::class,
```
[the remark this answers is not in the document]
Art4: This will allow us to have entries with `interface` => `classes` in the future. I have this use case while working on PSR-18 support.

src/Registry.php
```diff
@@ -112,24 +128,30 @@ class Registry
...
+
         if (!class_exists($class)) {
             return false;
         }
 
-        if (!((isset($this->legacyClassNames[$type]) && is_subclass_of($class, $type)) || (isset($this->default[$type]) && is_subclass_of($class, $this->default[$type])))) {
+        /** @var string */
+        $base_class = $this->default[$type] ?? $this->legacy[$type];
+
+        if (!@is_subclass_of($class, $base_class)) {
```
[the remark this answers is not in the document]
Art4: not sure. It was muted like this earlier. I plan to replace this with a `instanceof` check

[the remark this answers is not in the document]
Art4: Fixed.

src/Registry.php
```diff
@@ -136,27 +185,38 @@ public function register($type, $class, $legacy = false)
...
+     * @template T
+     * @param class-string<T> $type
+     * @return class-string<T>|null
      */
     public function get_class($type)
     {
-        if (!empty($this->classes[$type])) {
-            return $this->classes[$type];
+        if (array_key_exists($type, $this->legacyTypes)) {
+            // trigger_error(sprintf('"%s"(): Using argument #1 ($type) with value "%s" is deprecated since SimplePie 1.8, use class-string "%s" instead.', __METHOD__, $type, $this->legacyTypes[$type]), \E_USER_DEPRECATED);
```
mblaney, line 195: Do you want to remove or uncomment this error?

Art4: I would uncomment the trigger_error in the last minor release.

mblaney: If I merge this PR before the release it will be part of 1.8, so in that case the error seems applicable? Up to you though.

Art4: For consistency I would keep it as comment.

mblaney: ok I will merge this now then, I will wait for you to update the change log to include this PR before creating the release.

PR #768, opened 2022-12-31
No URL Decode for enclosure links

Alkarex:
Having an `urldecode` returns some potentially invalid URLs for enclosures, while SimplePie does not do that for other returned URLs.

Seems to be a similar error than https://github.com/simplepie/simplepie/commit/efb1d8e34c39276fb88fff0dcff01c673bec7e18

This line was introduced in https://github.com/simplepie/simplepie/commit/4bbd7eacfb6940d06899fa74a08754389ddd98de

Downstream PR https://github.com/FreshRSS/FreshRSS/pull/4944

P.S. There are a few other bugs related to enclosure links, and I will try to address some of them in distinct PRs.

### DIFF
--- a/tests/Unit/EnclosureTest.php
+++ b/tests/Unit/EnclosureTest.php
@@ -98,7 +98,8 @@ class EnclosureTest extends TestCase
         </item>
     </channel>
 </rss>
-EOT,
+EOT
+            ,
                 'http://example.net/link?a=%22b%22&amp;c=%3Cd%3E',
             ],
             'Test enclosure get_link urldecoded' => [
@@ -118,7 +119,8 @@ EOT,
         </item>
     </channel>
 </rss>
-EOT,
+EOT
+            ,
                 'http://example.net/link?a=%22b%22&amp;c=%3Cd%3E',
             ],
         ];
--- a/tests/Unit/EnclosureTest.php
+++ b/tests/Unit/EnclosureTest.php
@@ -83,7 +83,7 @@ class EnclosureTest extends TestCase
         return [
             'Test enclosure get_link urlencoded' => [
 <<<EOT
-<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:media="http://search.yahoo.com/mrss/">
+<rss version="2.0" xmlns:media="http://search.yahoo.com/mrss/">
 	<channel>
 		<title>Test enclosure link 1</title>
 		<description>Test enclosure link 1</description>
@@ -103,7 +103,7 @@ EOT,
             ],
             'Test enclosure get_link urldecoded' => [
 <<<EOT
-<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom" xmlns:media="http://search.yahoo.com/mrss/">
+<rss version="2.0" xmlns:media="http://search.yahoo.com/mrss/">
     <channel>
         <title>Test enclosure link 2</title>
         <description>Test enclosure link 2</description>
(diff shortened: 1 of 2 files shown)


Art4 (approved):

jtojnar:
> This line was introduced in [8a5f601#diff-a710c236cc7775672edbe5840d8fb96b0e91b0d56eff709f6eace2da47c1fe43R349-R359](https://github.com/simplepie/simplepie/commit/8a5f6013c4405cc8d5ccb27f2d76c78e8657df53#diff-a710c236cc7775672edbe5840d8fb96b0e91b0d56eff709f6eace2da47c1fe43R349-R359)

It has actually been introduced in 4bbd7eacfb6940d06899fa74a08754389ddd98de – one has to jump a bit since git blame is not perfect.

tests/Unit/EnclosureTest.php
```diff
@@ -58,4 +60,69 @@ public function testClassExists()
...
+            <media:content url="http://example.net/link?a=&quot;b&quot;&amp;c=&lt;d&gt;" medium="image">
+            </media:content>
+        </item>
+    </channel>
+</rss>
+EOT
+            ,
+                'http://example.net/link?a=%22b%22&amp;c=%3Cd%3E',
+            ],
+        ];
```
jtojnar, line 126: Might be a bit cleaner to switch to `yield` to reduce unnecessary indentation.

And switching the heredoc mark to `XML` will allow some editors (e.g. Sublime Text) to highlight it as such.

Ideally we would also move the ending mark to proper level but [flexible heredoc syntax](https://wiki.php.net/rfc/flexible_heredoc_nowdoc_syntaxes) is only available since PHP 7.3. Fortunately, this document is not whitespace sensitive so the extra indentation will not matter. We can move the marker once we bump the minimal PHP version.

```suggestion
        yield 'Test enclosure get_link urlencoded' => [
            <<<XML
            <rss version="2.0" xmlns:media="http://search.yahoo.com/mrss/">
                <channel>
                    <title>Test enclosure link 1</title>
                    <description>Test enclosure link 1</description>
                    <link>http://example.net/tests/</link>
                    <item>
                        <title>Test enclosure link 1.1</title>
                        <description>Test enclosure link 1.1</description>
                        <guid>http://example.net/tests/#1.1</guid>
                        <link>http://example.net/tests/#1.1</link>
                        <media:content url="http://example.net/link?a=%22b%22&amp;c=%3Cd%3E" medium="image">
                        </media:content>
                    </item>
                </channel>
            </rss>
XML
            ,
            'http://example.net/link?a=%22b%22&amp;c=%3Cd%3E',
        ];

        yield 'Test enclosure get_link urldecoded' => [
            <<<XML
            <rss version="2.0" xmlns:media="http://search.yahoo.com/mrss/">
                <channel>
                    <title>Test enclosure link 2</title>
                    <description>Test enclosure link 2</description>
                    <link>http://example.net/tests/</link>
                    <item>
                        <title>Test enclosure link 2.1</title>
                        <description>Test enclosure link 2.1</description>
                        <guid>http://example.net/tests/#2.1</guid>
                        <link>http://example.net/tests/#2.1</link>
                        <media:content url="http://example.net/link?a=&quot;b&quot;&amp;c=&lt;d&gt;" medium="image">
                        </media:content>
                    </item>
                </channel>
            </rss>
XML
            ,
            'http://example.net/link?a=%22b%22&amp;c=%3Cd%3E',
        ];
```

Alkarex: I had kept the same syntax than other tests, so for consistency, it could be an idea to update the rest.

Art4: If someone wants to do the work, then we can do that. On the other hand, we could also keep the old syntax for consistency and let php-cs-fixer fix it automatically once we have increased the PHP version.

PR #769, opened 2022-12-31
Fix case of multiple RSS2.0 enclosures

Alkarex:
SimplePie only returned the first enclosure when using the RSS 2.0 syntax for enclosures, instead of returning them all.

Please merge https://github.com/simplepie/simplepie/pull/768 first.

Downtream PR: https://github.com/FreshRSS/FreshRSS/pull/4944

### DIFF
--- a/src/Item.php
+++ b/src/Item.php
@@ -1,4 +1,6 @@
 <?php
+
+declare(strict_types=1);
 /**
  * SimplePie
  *
--- a/src/Item.php
+++ b/src/Item.php
@@ -2241,7 +2241,7 @@ class Item implements RegistryAware
                 }
             }
 
-            foreach ((array) $this->get_item_tags(\SimplePie\SimplePie::NAMESPACE_RSS_20, 'enclosure') as $enclosure) {
+            foreach ($this->get_item_tags(\SimplePie\SimplePie::NAMESPACE_RSS_20, 'enclosure') ?? [] as $enclosure) {
                 if (isset($enclosure['attribs']['']['url'])) {
                     // Attributes
                     $bitrate = null;
--- a/src/Item.php
+++ b/src/Item.php
@@ -370,7 +370,13 @@ class Item implements RegistryAware
     {
         if (!isset($this->data['thumbnail'])) {
             if ($return = $this->get_item_tags(\SimplePie\SimplePie::NAMESPACE_MEDIARSS, 'thumbnail')) {
-                $this->data['thumbnail'] = $return[0]['attribs'][''];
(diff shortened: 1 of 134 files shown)


tests/Unit/EnclosureTest.php
```diff
@@ -58,4 +60,135 @@ public function testClassExists()
...
+    {
+        $feed = new SimplePie();
+        $feed->set_raw_data($data);
+        $feed->enable_cache(false);
+        $feed->init();
+
+        $item = $feed->get_item(0);
+        $enclosures = $item->get_enclosures();
+
+        $this->assertSame($expected, count($enclosures));
```
Art4, line 142: What do you think about using `assetCount()`?

```suggestion
        $this->assertCount($expected, $item->get_enclosures());
```

src/Item.php
```diff
@@ -2241,8 +2241,8 @@ public function get_enclosures()
                 }
             }
 
-            if ($enclosure = $this->get_item_tags(\SimplePie\SimplePie::NAMESPACE_RSS_20, 'enclosure')) {
-                if (isset($enclosure[0]['attribs']['']['url'])) {
+            foreach ((array) $this->get_item_tags(\SimplePie\SimplePie::NAMESPACE_RSS_20, 'enclosure') as $enclosure) {
```
jtojnar, line 2244: I assume the cast is for the [`null` case](https://github.com/simplepie/simplepie/blob/839d34c10273f7eb2abe52fbb78a09d5727e14f9/src/Item.php#L152)?

If that is the case, something like the following might be less confusing.

```suggestion
            foreach ($this->get_item_tags(\SimplePie\SimplePie::NAMESPACE_RSS_20, 'enclosure') ?? [] as $enclosure) {
```

But maybe even better would be changing the `get_item_tags` to return empty array so that it actually respects the `@return` type annotation.

jtojnar:
It is not clear from the RSS spec if multiple enclosures are valid but it makes sense to me to err on the side of accepting it since the ambiguity allows feed producers to do this.

https://www.rssboard.org/rss-validator/docs/warning/DuplicateEnclosure.html

src/Item.php
Alkarex: Yes. I had kept the exact same syntax than lines above

Alkarex:
For the record, when using the `<media:content>` syntax, the specification explicitly accounts for multiple enclosures https://www.rssboard.org/media-rss#media-content
This PR adds a test for the  `<media:content>` syntax, while the fix is for the RSS 2.0 `<content>` syntax.

Art4 (approved):

PR #771, opened 2023-01-12
Fix broken phpdoc references

jtojnar:
Otherwise ApiGen will complain:

	Missing SimplePie\Parse\callback
	referenced by SimplePie\Parse\Date

	Missing SimplePie\Cache\type
	referenced by SimplePie\Cache\Redis

	Missing SimplePie\Cache\SimplePie
	referenced by SimplePie\Cache\Redis

	Missing SimplePie\Cache\PDO
	referenced by SimplePie\Cache\MySQL

### DIFF
--- a/src/Cache/Base.php
+++ b/src/Cache/Base.php
@@ -83,7 +83,7 @@ interface Base
     /**
      * Save data to the cache
      *
-     * @param array|SimplePie $data Data to store in the cache. If passed a SimplePie object, only cache the $data property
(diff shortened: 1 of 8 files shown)


Art4:
> Though we cannot do much about the PSR-16 symbols.

What about adding the `autoload.psr4` config to Simplepie src folder in `docs/apigen/composer.json`?

jtojnar:
> How about adding the `psr/simple-cache` dependency in `docs/apigen/composer.json`?

Yeah, that seems to work.

Issue #755, opened 2022-10-18
enable_exceptions does not work properly in all cases


jtojnar:
https://github.com/simplepie/simplepie/blob/9e9add3428ce86aede874bcf9a59c78e272f8dc1/src/SimplePie.php#L1313-L1321

Maybe we should convert setting `$this->error` to throwing an exception on failure everywhere internally, catch it at the top level in the public functions, and rethrow it if `enable_exceptions` is true. And deprecate the `enable_exceptions` method.

For 2.0, we could then remove the `try-catch` statements and the `enable_exceptions` method.

Issue #431, opened 2016-03-09
Question: Code still up to date?


JanPetterMG:
Are this repo still maintained, and the code still up to date?
````Latest release, Oct 2012````, very few changes made after that, and there is a lot of old issues and pull requests...

I would like to switch to SimplePie, as I don't see that many competitors around, but I'm not sure if that's a good or bad thing to do. I've been maintaining and developing my own custom parser based on what once in a time was the _FeedForAll RSS2SQL_ parser, but I don't want to replace it with something that is not maintained anymore...

I've seen some forks is still active, is one of these the way to go? or does it exist any "new" semi-official repo somewhere intended to be a continuation?

mblaney:
Hi @JanPetterMG I keep my fork up to date with the changes I'm making, depends if that suits what you want to use it for though. I would love to collaborate if you're interested.

talgalili:
@mblaney - do you know of other people actively maintaining a fork?

Thanks.

rmccue:
Sorry folks, I don't really have time to maintain this outside of security patches. However, I know there's a few people who are interested in continuing to maintain this. The state that I've left SP in is pretty unacceptable, so apologies to everyone using it.

@mblaney Happy to grant you with push access and all the bells-and-whistles if you'd like to help maintain it? (As a fellow Brisbanite, I instinctively trust you. ;) )

I'd like to add more people to help maintain it as well, since I don't have the resources to. I know @jgalea and team have been interested in this, and there's still lots of folks using SP, including WordPress. Ideally, 3-4 contributors could step up to help maintain the repo collaboratively.

The main thing I ask for any changes: **Don't break backwards compatibility** - BC must be maintained above (almost) everything else. Adding more stuff is fine, refactoring is fine, as is deprecating, but don't remove anything.

talgalili:
Thanks @rmccue !

ifsnop:
Please, don't let SP die!

On Fri, Mar 11, 2016 at 1:52 PM, Tal Galili <notifications@github.com>
wrote:

> Thanks @rmccue <https://github.com/rmccue> !
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/simplepie/simplepie/issues/431#issuecomment-195353193>
> .
>



-- 
-- Use of a keyboard or mouse may be linked to serious injuries or
disorders.
diego dot torres at gmail dot com - Madrid / Spain

mblaney:
[...]

PR #774, opened 2023-01-24
Split SimplePie\File class into HTTP client and response

Art4:
Hi 👋 

At the moment we have `SimplePie\File::__construct()`, where a request is handled using `curl`, `fsockopen` or `file_get_contents`. The response is saved in the class properties like `body`, `headers`, and so on.

This PR is a first step to allow other HTTP libraries in SimplePie. So this PR splits the File class into 2 parts:

1. HTTP client with a `request()` method. The client internally calls the `File` class like before.
2. The `File` class now implements a `Response` interface, that allows getting all the data like before through methods.

The concept and interfaces are interoperable with [PSR-18](https://www.php-fig.org/psr/psr-18/) and [PSR-7](https://www.php-fig.org/psr/psr-7/). The next step will be an implementation of `SimplePie\HTTP\Client` that works with a PSR-18 client under the hood.

refs #520


mblaney:
looks good @Art4, before I merge can someone confirm that the tests are running? I'm seeing `all passed` but when I click through there are no details.

Art4:
@mblaney Github seems to have some problems at the moment. Github Actions is experiencing degraded performance. See https://www.githubstatus.com/

grafik

src/File.php
```diff
@@ -251,6 +256,195 @@ public function __construct($url, $timeout = 10, $redirects = 5, $headers = null
...
+                    $header_lines = explode(',', $header);
+                    $header = [];
+                    foreach ($header_lines as $header_line) {
+                        $header[] = trim($header_line);
+                    }
+                }
+            }
+
+            $this->parsed_headers[strtolower($name)] = $header;
+        }
```
Art4, line 274: Having the headers parsed in the __construct() method may be the fastest method, but has some drawbacks in BC. I will rewrite this so it will work without the `$parsed_headers` attribute.

jtojnar:
I will try to review it later today.

Art4:
Thank you, I appreciate it. I just found another error and stabilized it with a test.

src/File.php
```diff
@@ -56,17 +58,20 @@
...
  */
-class File
+class File implements Response
 {
+    /** @var string Holds the final url after following possible redirects */
     public $url;
     public $useragent;
     public $success = true;
     public $headers = [];
+    public $parsed_headers = [];
```
Art4, line 68: I should remove this unused attribute.

Art4: Done.

src/HTTP/Client.php
```diff
@@ -0,0 +1,72 @@
...
+ * AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDERS
+ * AND CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+ * OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * @package SimplePie
+ * @copyright 2004-2022 Ryan Parman, Sam Sneddon, Ryan McCue
```
jtojnar, line 38: I wonder if this should not actually contain your name since you own copyright to this file.

https://github.com/simplepie/simplepie/issues/778

src/HTTP/Client.php
```diff
@@ -0,0 +1,72 @@
...
+ * @license http://www.opensource.org/licenses/bsd-license.php BSD License
+ */
+
+namespace SimplePie\HTTP;
+
+/**
+ * HTTP Client interface
+ *
+ * @package SimplePie
+ * @subpackage HTTP
```
jtojnar, line 54: [`subpackage`](https://docs.phpdoc.org/2.9/references/phpdoc/tags/subpackage.html) is deprecated and [`package`](https://docs.phpdoc.org/2.9/references/phpdoc/tags/package.html) is redundant.

src/HTTP/Client.php
```diff
@@ -0,0 +1,72 @@
...
+{
+    public const METHOD_GET = 'GET';
+
+    /**
+     * send a request and return the response
+     *
+     * @param string $method
+     * @param string $url
+     * @param array $headers in the form string[]
+     * @param array $options
```
jtojnar: No options argument. And no need to include the redundant params.

```suggestion
     * @param string[] $headers
```

src/HTTP/Client.php
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @return Response
```
jtojnar:
```suggestion
```

src/HTTP/Response.php
```diff
@@ -0,0 +1,199 @@
...
+/**
+ * HTTP Response interface
+ *
+ * This interface must be interoperable with Psr\Http\Message\ResponseInterface
+ * @see https://www.php-fig.org/psr/psr-7/#33-psrhttpmessageresponseinterface
+ *
+ * @package SimplePie
+ * @subpackage HTTP
+ */
+interface Response
```
jtojnar, line 57: Why not use `Psr\Http\Message\ResponseInterface`?

src/HTTP/Client.php
```diff
@@ -0,0 +1,72 @@
...
+namespace SimplePie\HTTP;
+
+/**
+ * HTTP Client interface
+ *
+ * @package SimplePie
+ * @subpackage HTTP
+ * @internal
+ */
+interface Client
```
jtojnar, line 57: Why not use `Psr\Http\Client\ClientInterface`?

src/SimplePie.php
```diff
@@ -1774,25 +1777,29 @@ protected function fetch_data(&$cache)
...
                             if ($this->force_cache_fallback) {
                                 $cache->set_data($cacheKey, $this->data, $this->cache_duration);
                                 return true;
                             }
 
                             unset($file);
+                        } finally {
+                            $this->timeout = $orig_timeout;
+                        }
+                        $this->status_code = $file->status_code;
```
jtojnar: Won’t `$file` be `null` when exception occurs and `$this->force_cache_fallback` is `false`?

src/SimplePie.php
```diff
@@ -1816,13 +1823,20 @@ protected function fetch_data(&$cache)
...
+                    $this->error = $th->getMessage();
+
+                    return !empty($this->data);
+                }
             }
         }
         $this->status_code = $file->status_code;
 
         // If the file connection has an error, set SimplePie::error to that and quit
-        if (!$file->success && !($file->method & self::FILE_SOURCE_REMOTE === 0 || ($file->status_code === 200 || $file->status_code > 206 && $file->status_code < 300))) {
```
jtojnar, line 1825: Same issue here.

src/Sanitize.php
```diff
@@ -166,6 +174,11 @@ public function pass_cache_data($enable_cache = true, $cache_location = './cache
         }
     }
 
+    public function set_http_client(Client $http_client): void
+    {
+        $this->http_client = $http_client;
+    }
+
     public function pass_file_data($file_class = 'SimplePie\File', $timeout = 10, $useragent = '', $force_fsockopen = false)
```
jtojnar, line 182: Perhaps we can deprecate this method now in favour of `set_http_client`.

src/Locator.php
```diff
@@ -226,9 +236,15 @@ protected function search_elements_by_tag($name, &$done, $feeds)
                     $headers = [
                         'Accept' => 'application/atom+xml, application/rss+xml, application/rdf+xml;q=0.9, application/xml;q=0.8, text/xml;q=0.8, text/html;q=0.7, unknown/unknown;q=0.1, application/unknown;q=0.1, */*;q=0.1',
                     ];
-                    $feed = $this->registry->create(File::class, [$href, $this->timeout, 5, $headers, $this->useragent, $this->force_fsockopen, $this->curl_options]);
-                    if ($feed->success && ($feed->method & \SimplePie\SimplePie::FILE_SOURCE_REMOTE === 0 || ($feed->status_code === 200 || $feed->status_code > 206 && $feed->status_code < 300)) && $this->is_feed($feed, true)) {
-                        $feeds[$href] = $feed;
+
+                    try {
+                        $file = $this->get_http_client()->request(Client::METHOD_GET, $href, $headers);
```
jtojnar: How do we deal with redirects?

src/Locator.php
```diff
@@ -71,14 +75,20 @@ class Locator implements RegistryAware
     public $dom;
     protected $registry;
 
-    public function __construct(\SimplePie\File $file, $timeout = 10, $useragent = null, $max_checked_feeds = 10, $force_fsockopen = false, $curl_options = [])
+    /**
+     * @var Client
+     */
+    private $http_client = null;
+
+    public function __construct(\SimplePie\File $file, $timeout = 10, $useragent = null, $max_checked_feeds = 10, $force_fsockopen = false, $curl_options = [], ?Client $http_client = null)
```
jtojnar, line 83: Ugly but I do not see a better way to make it preserve compatibility. Other than maybe using `func_get_args` and introspection. But that would be even uglier.

src/File.php
```diff
@@ -70,6 +71,7 @@ class File implements Response
     public $redirects = 0;
     public $error;
     public $method = \SimplePie\SimplePie::FILE_SOURCE_NONE;
+    /** @var string Holds the original requested URL */
```
jtojnar, line 73: That is not true. If a permanent redirect is encountered, it will hold the redirected URL:

https://github.com/simplepie/simplepie/blob/e597f27d414a774f732c0b260d67e11f6b6a7524/src/File.php#L139

src/HTTP/Client.php
Art4: To be honest I don't really care about having my name in a project I contribute. :relaxed: I would also rather see something like "SimplePie Team and contributers" than my name.

src/HTTP/Client.php
Art4: I know too little about phpdoc conventions. Can you make a suggestion how this should look like?

jtojnar: The [`subpackage`](https://docs.phpdoc.org/2.9/references/phpdoc/tags/subpackage.html) documentation suggests `@package SimplePie\HTTP`, and [`package`](https://docs.phpdoc.org/2.9/references/phpdoc/tags/package.html) docs say it is not necessary when it is the same as the namespace. I would just drop both.

src/Locator.php
```diff
@@ -164,8 +164,6 @@ public function is_feed($file, $check_html = false)
             }
 
             return in_array($sniffed, $mime_types);
-        } elseif ($file->method & \SimplePie\SimplePie::FILE_SOURCE_LOCAL) {
```
jtojnar, line 157: This changes the semantics,

jtojnar: Addressed later.

src/HTTP/Response.php
Art4: We cannot implement PSR-7 ResponseInterface, because this would require the existence of the interfaces, so we would have to include them into the build. But that could lead to errors if the user hat the interfaces already defined themself or have them installed via composer.

Also the `ResponseInterface` is to big for out use case. We would have to implement the `with*` setters and also `UriInterface` and `StreamInterface`. However, we do not need these (yet). So I defined the `Response` interface as a subset of the `ResponseInterface` like we have done it with the PSR-16 SimpleCache Interfaces. And at a last point please note, that the `Response` interface is marked as internal. So we could later change it or replace it with PSR-7 `ResponseInterface`.

I would recommend to take a look at the follow up PR #777 where I add support for PSR-7 responses.

src/HTTP/Client.php
Art4: Please see my comment about the Response interface: https://github.com/simplepie/simplepie/pull/774#discussion_r1089779057

src/Sanitize.php
Art4: this would be a topic for #777 where this method will be added.

src/Locator.php
Art4: The `File` class handles redirects in the constructor.

src/Locator.php
Art4: Another idea that I had would be to allow the storage of instances in the `Registry`. This way we could get the http client from the registry instead of setting it in the constructor. But that could be done in another PR.

src/File.php
Art4: Yes, that is the only exception. But we can also remove the comments. I only added them while understanding the meanings of them.

jtojnar: I think the comment is useful. The meaning of the property is not obvious and I have to check the code every time to make sense of it.

src/HTTP/Response.php
jtojnar: To be honest, I would just drop the compiled library, rather than create more unnecessary bridging classes, forever complicating the API. People should just use Composer these days. And if we really need to support that use case, we can always adjust the build to wrap third-party libraries inside `if (!class_exists('...'))`.

As for the `with*` “setters”, the `Response` object is supposed to be immutable (a good thing IMO, the repeated calls of constructor are a travesty) so the methods are necessary.

Art4: Dropping the compiled build could be made with release 2.0.0, but at the moment this file is used by some projects, see [here](https://github.com/LeedRSS/Leed/commit/9f2d924239a6e616d31cf47ff98a819652d686c0) or [here](https://github.com/ophian/styx/commit/13ff57d30f7647edcde0a630dfefde8664598f50). Sure, dropping BC will made refactore more easier, but I also see this as a sporting challenge. :blush: SimplePie is also used by projects like WordPress, so one requirement for changes is to keep BC, see https://github.com/simplepie/simplepie/issues/431#issuecomment-195352216

src/SimplePie.php
Art4: Yes, you are right. I will create a test for that.

src/HTTP/Client.php
Art4: Ok, if I understand you and the phpdoc documentation correct, we could remove all `@package` and `@subpackage` tags since we introduced the namespaces. Because this topic is unrelated to this PR I would recommend to move theses changes into a separate PR. I've created #780.

src/File.php
Art4: I fixed the comment.

src/HTTP/Client.php
Art4: Done.

src/HTTP/Client.php
Art4: Done.

src/SimplePie.php
Art4: I think that everything is right here.
1. If `$file` isn't set, but `$this->file` exists, then `$file =& $this->file;` will be used.
2. If `$file` isn't set and `$this->file` doesn't exist, then `$file` will be set by `$file = $this->get_http_client()->request(...)`.
3. And if `$this->get_http_client()` throws an exception than the method will return `$this->data` and `$file` is not needed at all.

src/SimplePie.php
Art4: It was not possible for me to create a test for this complicated place at the moment, but I made sure $this->status_code is always set.

Art4:
After merge of #781 I will remove the `@package` tags in the new code.

src/HTTP/Response.php
jtojnar: > Dropping the compiled build could be made with release 2.0.0, but at the moment this file is used by some projects, see [here](https://github.com/LeedRSS/Leed/commit/9f2d924239a6e616d31cf47ff98a819652d686c0) or [here](https://github.com/ophian/styx/commit/13ff57d30f7647edcde0a630dfefde8664598f50).

I asked about it in https://github.com/simplepie/simplepie/issues/782.

> My goal is to deprecate the whole `File` class, and I try to touch it as little as possible. Same with the IRI class and some other classes that would be obsolete with PSR-18 support.

Would these bridge classes be dropped in favour of direct PSR-18 as well? It does not look like it from #777

> Sure, dropping BC will made refactore more easier, but I also see this as a sporting challenge. blush SimplePie is also used by projects like WordPress, so one requirement for changes is to keep BC, see [#431 (comment)](https://github.com/simplepie/simplepie/issues/431#issuecomment-195352216)

You cannot really have software that is not frozen without BC break, see [Hyrum’s law](https://www.hyrumslaw.com/) and [TeX versioning](https://en.wikipedia.org/wiki/TeX#TeX82). At best, you can select a surface that you will aim to remain backwards compatible.

For example, let’s say I have a helper function `private function get_headers(Response $response): array` in my `File` subclass. Updating to this PR will make my app crash with `Fatal error: Access level to MyFileClass::get_headers() must be public (as in class File) `.

Or to bring another example, the recent minimum PHP version bump.

Each extra class and line of code has maintenance costs so I would prefer if we did not introduce new ones unless the benefits outweigh the costs.

And I want to argue availability of a bundle should not be part of the frozen surface, which would allow us less indirections here.

Art4: > Would these bridge classes be dropped in favour of direct PSR-18 as well? It does not look like it from #777

No. Dropping classes is a breaking change. Direct use of PSR-18 (and dropping the bridge classes) can be happen earliest with release 2.0.0.

> You cannot really have software that is not frozen without BC break, see [Hyrum’s law](https://www.hyrumslaw.com/) and [TeX versioning](https://en.wikipedia.org/wiki/TeX#TeX82). At best, you can select a surface that you will aim to remain backwards compatible.

Yes, and for the selection of this surface I recommend the Symfony BC promise: https://symfony.com/doc/current/contributing/code/bc.html
For your example with the new private helper method in an extended class there is no BC promise.

Art4: To be clear about the BC: For me there is no problem introducing breaking changes and releasing a new major version of SimplePie every few month. But the BC wish came from @rmccue, see https://github.com/simplepie/simplepie/issues/431#issuecomment-195352216

> The main thing I ask for any changes: **Don't break backwards compatibility** - BC must be maintained above (almost) everything else. Adding more stuff is fine, refactoring is fine, as is deprecating, but don't remove anything.

While working on the refactoring I'm trying to serve this wish as best as I can. I think even after release of SimplePie 2 there will be need for backports in the latest `one-dot-x` branch for some time,

jtojnar:
Sorry, forgot to send this. Continuing from https://github.com/simplepie/simplepie/issues/782#issuecomment-1412204943

> But: To ensure BC in version 1.x without adding requirements to external interfaces/libraries, a reasonable way for me is to create new layers that allow alternative libraries and deprecate old classes. The layers would also not mean "duplicating the interfaces", but allows us to prune the interfaces to a small subset that we need. We can even change or throw them away at any time because they are marked as `@internal` and thus exclude from the BC promise.

Yes, it is a reasonable approach if we do not want to introduce dependencies but it is not clear to me that this is a constraint that we have to uphold or that we should. Maintaining the [Symfony promise](https://symfony.com/doc/current/contributing/code/bc.html) sounds sufficientl reasonable to me and I believe we could have do that with much less code to maintain. Duplicating and pruning is just duplicating with an extra step. The maintenance concerns still apply. We cannot really mark these interfaces like `HTTP\Client` from #774 internal since projects would be meant to implement them.

>   I definitely prefer implementing our own interfaces/layers over implementing the PSR interfaces. In the words of PSR-18 SimplePie is a "[calling library](https://www.php-fig.org/psr/psr-18/#definitions)" and therefor "does not implement this specification's interfaces but consumes an object that implements them (a Client)."

But SimplePie also contains a client (`File`). If we implemented the interfaces for it, we could mostly just accept PSR-18 interfaces for it without any extra layers.

> For me a deprecation means that I don't have to actively care about it anymore (as long as all tests work). **If users report a bug in the deprecated classes, I would point out the alternative or ask for a PR. I would not allow to add new features to deprecated classes** to increase the pressure on switching to the alternative.

Sure but after removal of `FileClient` we would have two options:

1. Keep the `Client` interface forever with a single `Psr18Client`, thus having an extra useless layer
2. Deprecate `Client` interface and the extra classes and have SimplePie accept `Psr\Http\Client\ClientInterface` directly.

From what you posted, I understand you want the latter.

But then we will have a codebase in a state where we accept both native PSR objects and the interop objects, which is not much different from accepting both legacy `File` class and native PSR objects I am suggesting so we could skip the interop objects altogether. (Not sure whether to implement PSR-18 in `File`, that would require resolving #782, or temporarily have ugly `if`s throughout the codebase.)

jtojnar:
Reading the comments bit more, I now realize the confusion might have come from my assumption that there would be a deprecation period for the interop layers. If Scenario A removes them without prior deprecation, the extra code might be worth it for a slightly cleaner implementation on SimplePie side before the removal (no need for ugly `if`s throughout the codebase). But the downside is that consumers would have to migrate twice (custom `File` → `Psr18Client` → native PSR Client) so not sure that BC perspective can really justify that.

Art4:
> Yes, it is a reasonable approach if we do not want to introduce dependencies but it is not clear to me that this is a constraint that we have to uphold or that we should. Maintaining the [Symfony promise](https://symfony.com/doc/current/contributing/code/bc.html) sounds sufficientl reasonable to me and I believe we could have do that with much less code to maintain. 

Thank you for your comment. As I explained in #782 this PR was made under the assumption that new external dependencies will be a BC break (not allowed in 1.x, but allowed in 2.x), so I'm with you that directly require PSR-17/PSR-18 implementations would require much less code. This decision is up to @mblaney and the team and would impact #777.

> Duplicating and pruning is just duplicating with an extra step. The maintenance concerns still apply. 
> [...]
> But SimplePie also contains a client (File). If we implemented the interfaces for it, we could mostly just accept PSR-18 interfaces for it without any extra layers.

I'm not sure if I unterstand you correctly. (I do not want to exclude the possibility that I fail in understanding because of the language barrier.) Do you want to implement (some of) the PSR-7/PSR-18 interfaces directly in/with `SimplePie\File`? Why?

I've implemented PSR-7, PSR-17 and PSR-18 already a few times in different projects. The last time was for [Requests](https://github.com/WordPress/Requests/pull/768), which then resulted in its [own project](https://github.com/Art4/WP-Requests-PSR18-Adapter).

Implementing these PSRs in SimplePie would require us to implement 12 interfaces with 57 methods.

PSR | Interface | Methods
-- | -- | --
PSR-7 | MessageInterface | 11
  | RequestInterface | 6
  | ResponseInterface | 3
  | StreamInterface | 15
  | UriInterface | 16
PSR-17 | RequestFactoryInterface | 1
  | StreamFactoryInterface | 3
  | UriFactoryInterface | 1
PSR-18 | ClientInterface | 1
  | ClientExceptionInterface | 0
  | RequestExceptionInterface | 0
  | NetworkExceptionInterface | 0
Summary | 12 | 57

For the interop layer in SimplePie I condensed them to 3 interfaces with 9 methods. And they will be removed again in SimplePie 2.

Namespace | Interface | Methods
-- | -- | --
SimplePie\Exception | HttpException | 0
SimplePie\HTTP | Client | 1
  | Response | 8
Summary | 3 | 9

I cannot understand your point why maintaining this implementation concerns you more than maintaining a PSR implementation?

> We cannot really mark these interfaces like `HTTP\Client` from #774 internal since projects would be meant to implement them.

Yes, you are right. 👍 I just realized that we cannot accept the internal `SimplePie\HTTP\Client` as parameter in `Locator::__construct()` or `Sanitize::set_http_client()`. Instead we must accept PSR-18 client and PSR-17 factories like in #777. I will remove the `@internal` for now from `SimplePie\HTTP\Client` but will fix this in #777 again to keep this PR small.

> Sure but after removal of `FileClient` we would have two options:
> 
> 1. Keep the `Client` interface forever with a single `Psr18Client`, thus having an extra useless layer
> 2. Deprecate `Client` interface and the extra classes and have SimplePie accept `Psr\Http\Client\ClientInterface` directly.
> 
> From what you posted, I understand you want the latter.

No, I prefer option 3 (for SimplePie 2):

3. _Remove_ `Client` interface and the extra classes and have SimplePie using `Psr\Http\Client\ClientInterface` directly.

That's why I want to keep `SimplePie\HTTP\Client` internal. This code is only a Forward compatability layer to allow a smoother upgrade. I want to delete it in SimplePie 2 again.

src/HTTP/Response.php
```diff
@@ -0,0 +1,197 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SimplePie
```
jtojnar, line 5: Since this is based on PSR code, we need to attribute it properly and mention the license:

https://github.com/php-fig/http-message/blob/efd67d1dc14a7ef4fc4e518e7dee91c271d524e4/LICENSE#L1

This was actually one of the reasons I suggested SPDX, then we could just plop at the top:

```php
// SPDX-FileCopyrightText: 2014 PHP Framework Interoperability Group
// SPDX-FileCopyrightText: 2023 Art4 <art4@wlabs.de>
// SPDX-License-Identifier: MIT
```

We would also need to include the MIT license file (perhaps copy it into the repo as `LICENSE.MIT.md`. We could remove it once we remove this compat code.

src/Exception/HttpException.php
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @copyright 2004-2022 Ryan Parman, Sam Sneddon, Ryan McCue
+ * @author Ryan Parman
+ * @author Sam Sneddon
+ * @author Ryan McCue
+ * @link http://simplepie.org/ SimplePie
+ * @license http://www.opensource.org/licenses/bsd-license.php BSD License
+ */
+
+namespace SimplePie\Exception;
```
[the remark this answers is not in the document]
Art4: I have no strong opinions about the place for exceptions. We will potentially get more Exceptions while fixing #755.

src/HTTP/Response.php
Art4: I would keep this for know to #798 as long as no decision has been made on this.

Art4:
@mblaney Is there any chance that this will be merged soon?

jtojnar:
I am still reviewing this.

Since my head is too small to look at everything at once, I am extracting some of the unrelated cleanups into separate pull requests and rebased and squashed the fixes into the original commits. I have also added some notes about the changes into the commit messages:

https://github.com/simplepie/simplepie/compare/master...jtojnar:simplepie:split-file-into-client-and-response

I have also switched to SPDX syntax since the long comments make it difficult to fit multiple files on the screen at the same time.

jtojnar (commented):
Sorry it took so long, I have finally managed to review this in detail it deserves.

You did a great job with this and convinced me that this approach is a reasonable interim solution.

While reviewing, I did some rebase surgery to make it easier to handle with my attention span, and while at it fixed some issues. You can see that in #815.

src/SimplePie.php
```diff
@@ -1778,25 +1781,30 @@ protected function fetch_data(&$cache)
...
-                                // is still valid.
-                                $this->raw_data = false;
-                                $cache->set_data($cacheKey, $this->data, $this->cache_duration);
-                                return true;
-                            }
-                        } else {
+                        try {
+                            $orig_timeout = $this->timeout;
+                            $this->timeout = 1;
+                            $file = $this->get_http_client()->request(Client::METHOD_GET, $this->feed_url, $headers);
```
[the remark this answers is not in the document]
Art4: This request will check if the cache has been changed. The timeout was to 1 sec. I assume that the intention was to not wait to long for a response for this check.

It then was changed to `$this->timeout/10` in https://github.com/simplepie/simplepie/commit/fe7e33301e0ac1df982b4652b4b2ae3372e2d254, but this could lead to a float instead of int, if e.g. `$this->timeout` is set to `5`. That's why I changed it back to 1 sec. Alternative we could use `ceil()` but this will be in most cases `1` or (if $this->timeout = 11-20) `2`. So I decided to use `1` directly.

About the side-channel signalling I decided to use this instead of a new argument because the timeout ist moved to PSR-18 HTTP client and  `$this->timeout` will be deprecated.

Art4:
@mblaney There are a couple of RRs that depend on this PR and #777 and it is a bit frustrating to fix resolve conflicts without getting feedback. Could you please review this PR and #777 first before other things?

src/SimplePie.php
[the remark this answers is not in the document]
jtojnar: Thanks for the context. Peeking further into history, it looks like it was `1` since the introduction of timeouts along with `SimplePie_File` in https://github.com/simplepie/simplepie/commit/5bf1814b6943be7708cf7689f01c1c6aeb901afe.

But I think it is still useful to keep it proportional to the master timeout to allow people with slow connection to increase it. Even if we assume majority of cached resources will not get outdated, waiting for server to reply `HTTP 304 Not Modified` can still take time (domain name resolution can be particularly slow). And if it turns out that the cached version is outdated, the web server will try to send an updated version. Having too low timeout can mean the request will be terminated, only to be started again with the original timeout.

So I do not actually think reducing the timeout makes much sense at all. The only explanation I can come up with is that it prevents essentially doubling the timeout when fetching a cached resource from an unresponsive server without `force_cache_fallback`  due to re-running the failed request. 
I have split the change to a separate commit in https://github.com/simplepie/simplepie/pull/815, including the explanation.

jtojnar:
I propose we go with https://github.com/simplepie/simplepie/pull/815 instead. This change set is complex enough to warrant not squashing refactorings like the switch to Response methods into functional changes like the actual porting to `FileClient`. Hopefully, we (or our descendants) can have a clean `git blame` in the future.

mblaney:
> @mblaney There are a couple of RRs that depend on this PR and #777 and it is a bit frustrating to see other PRs merged and resolve conflicts without getting feedback. Could you please review this PR and #777 first before other things?

sorry @Art4 you're right, this PR should've been dealt with before merging smaller patches. I'm happy with it so didn't need to comment, but was waiting for approval from @jtojnar before merging. I'll wait for the two of you to decide how to continue before merging anything else.

Art4:
> sorry @Art4 you're right, this PR should've been dealt with before merging smaller patches. I'm happy with it so didn't need to comment, but was waiting for approval from @jtojnar before merging. I'll wait for the two of you to decide how to continue before merging anything else.

I just wanted to kindly let you know that I have already shared all my thoughts and comments in this and other PRs/issues. I believe I have covered everything that needs to be said. I think it's now time for you to make a decision. Once you have made a decision, I can move forward with the next steps in #731. Thank you so much for your time and consideration.

jtojnar:
@Art4 Just to be clear, are you okay with mergin https://github.com/simplepie/simplepie/pull/815 in stead of this PR?

mblaney:
thanks for the update @Art4, as the person driving this project I'm happy for you to make the decisions! :-)

I'm not sure a modified PR was necessary, but happy to merge #815 instead since that has a few improvements.

src/SimplePie.php
```diff
@@ -3226,6 +3261,23 @@ private function get_cache(string $feed_url = ''): DataCache
 
         return $this->cache;
     }
+
+    /**
+     * Get a HTTP client
+     */
+    private function get_http_client(): Client
```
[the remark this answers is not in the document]
Art4: It is not worth it. This method is private and will be removed in v2 again. Moving this to Registry will lead to more BC issues.

Art4:
@jtojnar I think I have addressed all the comments relevant to this PR. Some issues in this PR are addressed in #777, because they can't be fixed here. All open comments are either "not important for now", "fine for now" or you have already suggested improvements in #815. As long as this is not explicitly requested, I don't see the need to replicate the improvements here, but would rather discuss them in #815 (and yes, I still have some thoughts on them). The discussion is otherwise even more scattered than it already is because of the many open issues.

@mblaney This PR is the base of or mentioned by #777, #813, #815, #812 and #810. These issues are partly the basis for further issues like #826 and #833. Closing this PR will be quite confusing in retrospect. 

So if you are happy with this PR for the most part, I recommend that we merge this PR as a base and discuss the other changes in the other PRs and issues.

I would also like to see a decision from you in #782 because it has far reaching meanings on issues like #777. And I'd appreciate any thoughts or reactions from you to #731.

jtojnar:
> @mblaney This PR is the base of or mentioned by #777, #813, #815, #812 and #810. These issues are partly the basis for further issues like #826 and #833. Closing this PR will be quite confusing in retrospect.

I would argue that having a clean commit history with the details in commit messages (as in #815) is more important than GitHub issues, as it facilitates git archaeology. Here, we at most get a single big squashed commit without any rationale for the changes, which will be more confusing when running `git blame`.

Additionally, the git repo is a more permanent record – it already survived move from the previous hosting provider while the issues did not.

You could also the  contents of this PR with commits from #815 using the following:

```
git fetch git@github.com:jtojnar/simplepie.git split-file-into-client-and-response
git reset --hard FETCH_HEAD
```

That would allow us to merge this PR (#774) with a cleaner and more descriptive commit history.

(And then the same thing with #777 and #826).

mblaney:
Thanks @Art4 and @jtojnar for the updates. I appreciate yours thoughts on commit history @jtojnar, but I would prefer authors of a PR completed their work, so will retract my earlier comment and now merge #774. I hope that's ok, if you want to open a new PR or modify #815  for any of the changes you've made it would be good to discuss them separately.

jtojnar:
@mblaney

> I would prefer authors of a PR completed their work

That would be nice but the quality of the software should not suffer for it. Building software is a collaborative effort anyway, so it should not be much of a problem, as long as authorship is properly credited.

Unfortunately, as it is now the history is a complete mess making any debugging using `git blame` and `git bisect` much more annoying than it needs to be. This is not much of a problem now but will be in the future, once we are trying to find out about long forgotten design decisions and get commits like https://github.com/simplepie/simplepie/commit/5bf1814b6943be7708cf7689f01c1c6aeb901afe

And even worse, there are also correctness errors in this PR (e.g. https://github.com/simplepie/simplepie/pull/774#discussion_r1141241016). IMO the `master` branch should not be merged into known broken state with the plan to fix it later. That again will break `git bisect`.

PR #775, opened 2023-01-24
Add type hints in all classes

### LINKED ISSUE #773: [BUG] Uncaught TypeError: array_slice(): Argument #3 ($length) must be of type ?int
Version: 1.8

#### Description

WackoWiki feed action:
`$feed->get_items($pagination['offset'], $max)`

PHP Fatal error:  Uncaught TypeError: array_slice(): Argument #3 ($length) must be of type ?int, string given in /lib/SimplePie/src/SimplePie.php:30..

#### Suggestion

`$feed->get_items($pagination['offset'], (int) $max)`

Maybe cast $end with (int) also in array_slice?

`return array_slice($items, $start, (int) $end);`

And check example code with `get_items()`.

Art4:
This PR adds type hints to all methods in the `SimplePie` class where the arguments are documented. Thanks to type widening in PHP 7.2 this is not a breaking change.

There are also many methods where the parameters are not documented. This places should be be documented, but for BC they should trigger a deprecation warning, if a parameter has to be casted. The type declaration could be added in SimplePie 2.0.0 with a breaking change.

This will fix #773.

### DIFF
--- a/src/Enclosure.php
+++ b/src/Enclosure.php
@@ -294,7 +294,7 @@ class Enclosure
      * @param int $key
      * @return \SimplePie\Caption|null
      */
-    public function get_caption($key = 0)
+    public function get_caption(int $key = 0)
     {
         $captions = $this->get_captions();
         if (isset($captions[$key])) {
@@ -324,7 +324,7 @@ class Enclosure
      * @param int $key
      * @return \SimplePie\Category|null
      */
-    public function get_category($key = 0)
+    public function get_category(int $key = 0)
     {
         $categories = $this->get_categories();
         if (isset($categories[$key])) {
@@ -382,7 +382,7 @@ class Enclosure
      * @param int $key
      * @return \SimplePie\Credit|null
      */
-    public function get_credit($key = 0)
+    public function get_credit(int $key = 0)
     {
         $credits = $this->get_credits();
         if (isset($credits[$key])) {
@@ -426,7 +426,7 @@ class Enclosure
      * @param bool $convert Convert seconds into hh:mm:ss
      * @return string|int|null 'hh:mm:ss' string if `$convert` was specified, otherwise integer (or null if none found)
      */
-    public function get_duration($convert = false)
+    public function get_duration(bool $convert = false)
     {
         if ($this->duration !== null) {
             if ($convert) {
@@ -501,7 +501,7 @@ class Enclosure
      * @param int $key
      * @return string|null Hash as per `media:hash`, prefixed with "$algo:"
      */
-    public function get_hash($key = 0)
+    public function get_hash(int $key = 0)
     {
         $hashes = $this->get_hashes();
         if (isset($hashes[$key])) {
@@ -560,7 +560,7 @@ class Enclosure
      * @param int $key
      * @return string|null
      */
-    public function get_keyword($key = 0)
+    public function get_keyword(int $key = 0)
     {
         $keywords = $this->get_keywords();
         if (isset($keywords[$key])) {
@@ -648,7 +648,7 @@ class Enclosure
      * @param int $key
      * @return \SimplePie\Rating|null
      */
-    public function get_rating($key = 0)
+    public function get_rating(int $key = 0)
     {
         $ratings = $this->get_ratings();
         if (isset($ratings[$key])) {
@@ -678,7 +678,7 @@ class Enclosure
      * @param int $key
      * @return \SimplePie\Restriction|null
      */
-    public function get_restriction($key = 0)
+    public function get_restriction(int $key = 0)
     {
         $restrictions = $this->get_restrictions();
         if (isset($restrictions[$key])) {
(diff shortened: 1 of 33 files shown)


jtojnar (commented):
Looks good.

We can get even more with `'phpdoc_to_param_type' => true` rule in PHP-CS-Fixer.

Art4:
Didn't know about this rule. I will try it. Thank you. 👍

jtojnar:
Would be nice to be able to use `phpdoc_to_return_type` as well but we would need to limit it to just private functions (something like https://fekir.info/post/automate-git-interactive-add/ would allow this easily).

Art4:
> Would be nice to be able to use `phpdoc_to_return_type` as well but we would need to limit it to just private functions (something like https://fekir.info/post/automate-git-interactive-add/ would allow this easily).

I've added them manually. We don't have much private methods. 🙂

src/Enclosure.php
```diff
@@ -1115,7 +1115,7 @@ public function get_real_type($find_handler = false)
                     $type = 'audio/x-ms-wma';
                     break;
 
-                // Video mime-types
+                    // Video mime-types
```
jtojnar, line 1118: Weird indent, php-cs-fixer bug?

src/Item.php
```diff
@@ -905,7 +905,7 @@ public function get_links($rel = 'alternate')
      * @param int $key The enclosure that you want to return.  Remember that arrays begin with 0, not 1
      * @return \SimplePie\Enclosure|null
      */
-    public function get_enclosure($key = 0, $prefer = null)
+    public function get_enclosure(int $key = 0, $prefer = null)
```
jtojnar, line 908: `$prefer` is a weird argument introduced in 4bbd7eacfb6940d06899fa74a08754389ddd98de, probably can be safely removed, it has never been used according to `git log -p -G 'get_enclosure\b.+,'` and `git log -p -G '\$prefer'`.

src/Misc.php
```diff
@@ -1746,7 +1746,7 @@ public static function entities_decode($data)
     /**
      * Remove RFC822 comments
      *
-     * @param string $data Data to strip comments from
+     * @param string $string Data to strip comments from
      * @return string Comment stripped string
      */
     public static function uncomment_rfc822($string)
```
jtojnar, line 1752: Can add typehint here?

src/Misc.php
```diff
@@ -1904,7 +1904,7 @@ public static function space_separated_tokens($string)
      * @param int $codepoint Unicode codepoint
      * @return string UTF-8 character
      */
-    public static function codepoint_to_utf8($codepoint)
+    public static function codepoint_to_utf8(int $codepoint)
     {
         $codepoint = (int) $codepoint;
```
jtojnar, line 1909: Unneeded cast.

src/Enclosure.php
Art4: Yes, this is a bug in php-cs-fixer. I move the comment under the case statement.

src/Misc.php
Art4: Done.

src/Misc.php
Art4: I removed the type cast.

src/Misc.php
```diff
@@ -1749,7 +1749,7 @@ public static function entities_decode(string $data)
      * @param string $string Data to strip comments from
      * @return string Comment stripped string
      */
-    public static function uncomment_rfc822($string)
+    public static function uncomment_rfc822(string $string)
     {
         $string = (string) $string;
```
jtojnar, line 1754: Another cast that can be dropped.

src/Parse/Date.php
```diff
@@ -662,7 +662,7 @@ public function parse($date)
      * @access public
      * @param callable $callback
      */
-    public function add_callback($callback)
+    public function add_callback(callable $callback)
     {
         if (is_callable($callback)) {
```
jtojnar, line 667: Redundant check.

src/Item.php
```diff
@@ -647,7 +647,7 @@ public function get_copyright()
      * @param string $date_format Supports any PHP date format from {@see http://php.net/date} (empty for the raw data)
      * @return int|string|null
      */
-    public function get_date($date_format = 'j F Y, g:i a')
+    public function get_date(string $date_format = 'j F Y, g:i a')
```
jtojnar, line 650: Apparently, this [accepted false](https://github.com/simplepie/simplepie/commit/4dac51faf47780847f32820b24c4b573247f9ca4) in the past but with the typehint it will still work as before since without strict type, it will be autocast. But we can drop the two casts below.

Art4: The removal for `false` support was made in v1.3 which was technically a breaking change. But now it is documented and I would argue that adding the typehint would be ok.

src/Parse/Date.php
Art4: I removed the redundant check.

src/Item.php
Art4: I removed the autocasting.

jtojnar (approved):

Issue #684, opened 2021-06-10
[ENH] Consider better mf2 support


janboddez:
####  Is your feature request related to a problem? Please describe.

SimplePie takes Mf2's output but disregards most of it. It also sets `description` based on, e.g., `e-content` but not, let's say, `p-summary`, which would be a much better fit for `description`.

#### Describe the solution you'd like

1. Improve the algorithm that sets `content` and `description` for mf2 feeds.

2. A `get_data()` method that returns the complete Mf2 data array would also be nice, and is very easy to implement. It'd allow SimplePie clients to do a whole lot more than is possible now (i.e., use the information in the `reply-to` or `bookmark-of` or `repost-of` arrays, most of which, again, is currently disregarded).

#### Describe alternatives you've considered

Was thinking along the lines of making `parse_microformats()` protected rather than private, so that software that integrates SimplePie can more easily override it.

Then I found `set_parser_class()` so I think I might be able to work off that.

Not sure (this would be my first SimplePie project) if this could eventually be spin off into a plugin.

PR #776, opened 2023-01-25
Replace idna_convert.class.php with Algo26\IdnaConvert, allow custom idna library

### LINKED ISSUE #538: bundled idn/idna_convert.class.php can be replaced with mso/idna-convert
https://github.com/simplepie/simplepie/blob/1.5/idn/idna_convert.class.php

this class official upstream is:
- packagist: https://packagist.org/packages/mso/idna-convert
- github: https://github.com/phlylabs/idna-convert

if you have any local changes those should be reported first upstream:
- https://github.com/phlylabs/idna-convert/issues
- https://github.com/phlylabs/idna-convert/pulls

Art4:
Hi all 👋 

In #538 we discussed the update or replacement of idna_convert.class.php 0.5.1, which is outdated ~and probably don't work with PHP 7 or PHP 8 (at least I couldn't make it run)~. (I made it run. \o/)

This PR ~will remove the outdated bundled `idna_convert.class.php` file and~ will add support for the updated `Algo26\IdnaConvert` library. This PR fixes #538.

### Support for Algo26\IdnaConvert v2 and v3

The `idna_convert.class.php` has been updated and renamed to `Algo26\IdnaConvert`, see https://github.com/algo26-matthias/idna-convert
This PR will allow to use version 2 or 3. If the library is installed, it will work out of the box without other configuration. All you need to do is to install the library e.g. with composer:

```
$ composer require algo26-matthias/idna-convert:^3
```

I've added the library to the suggest section in the `composer.json`.

### BC: Support for idna_convert v0.5.1

If you would like to use the integrated idna_convert class (v0.5.1) you have to `require()` the file like before:

```php
require_once($path_to_simplepie_folder . '/idn/idna_convert.class.php');
$simplepie = new \SimplePie\SimplePie();
// ...
```

This PR will add more jobs in Github Actions to test SimplePie with the `Algo26\IdnaConvert` library v2 and v3 and v0.5.1 installed.

### Support for custom Idna converter

Also this PR adds a new `IdnaDomainFilter` interface. This will allow us to use a complete other solution by providing an implementation to the `Registry`.

```php
use SimplePie\Idna\IdnaDomainFilter;

class MyIdnaConverter implements IdnaDomainFilter {
    public function filterDomain(string $decoded): string {
        // do your own encoding here...
        return \idn_to_ascii($decoded);
    }
}

$simplepie = new \SimplePie\SimplePie();
$simplepie->get_registry()->register(IdnaDomainFilter::class, MyIdnaConverter::class);
```

### Follow up

This PR requires #774 to merge, so I can move the idna support from the `File` class to the HTTP client. ~On the other hand idna support is broken for PHP 7, so it would possibly ok to just remove it from the File class.~

In follow up I plan to use the same concept to improve the mf2 support (#684) and using [Requests](https://github.com/WordPress/Requests) as HTTP library (#520).


Art4:
Reading [this comment](https://github.com/simplepie/simplepie/issues/431#issuecomment-195352216) from @rmccue:

> The main thing I ask for any changes: Don't break backwards compatibility - BC must be maintained above (almost) everything else. Adding more stuff is fine, refactoring is fine, as is deprecating, but don't remove anything.

I think I should restore the `idna_convert.class.php` file and make it work to keep BC.

tests/Integration/IdnaSupportTest.php
```diff
@@ -0,0 +1,84 @@
...
+ * @license http://www.opensource.org/licenses/bsd-license.php BSD License
+ */
+
+namespace SimplePie\Tests\Integration;
+
+use Algo26\IdnaConvert\IdnaConvert;
+use Algo26\IdnaConvert\ToUnicode;
+use PHPUnit\Framework\TestCase;
+
+class IdnaSupportTest extends TestCase
```
jtojnar: It is not clear to me what the point of this test is. This should be already tested by `idna-convert` itself.

Art4: This will be addressed in the later commits. I created the needed functionality in this test and later moved it into the SimplePie\Idna\IdnaConverter class.

src/Enclosure.php
```diff
@@ -255,12 +258,32 @@ public function __construct($link = null, $type = null, $length = null, $javascr
...
+     * Set the Registry into the class
+     *
+     * @param Registry $registry
+     *
+     * @return void
+     */
+    public function set_registry(Registry $registry): void
+    {
+        $idnaConverter = $registry->create(IdnaDomainFilter::class);
+        $parsed = $registry->call(Misc::class, 'parse_url', [$this->link]);
```
jtojnar: Why replace a simple pure function like this with indirection via registry? It is unlikely that we would want to swap it and it just makes the code more complicated.

Art4: Yeah, you are right, but this is the only possibility to mock the Misc class for tests.

We should discuss the whole concept of the Registry in a separate issue and how we should refactore it.

Art4:
I've created BC support for idna_convert v0.5.1 and have added new tests in Github Actions to test the bundled library with PHP 8.2.

.github/workflows/ci.yml
```diff
@@ -87,7 +87,7 @@ jobs:
 
       - name: Run unit tests with idna_convert 0.5.1
         if: ${{ matrix.with == 'idna-convert-v051' }}
-        run: composer test --bootstrap=tests/bootstrap-with-idna-convert-051.php
+        run: bin/phpunit --bootstrap=tests/bootstrap-with-idna-convert-051.php
```
jtojnar, line 91: The proper syntax is `composer test -- --bootstrap=tests/bootstrap-with-idna-convert-051.php`

The `--` will stop composer from interpreting arguments as flags and let them pass through.

.github/workflows/ci.yml
```diff
@@ -62,7 +62,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
```
jtojnar, line 66: This should probably go to a separate PR.

Art4: I don't think this is a big thing. I just wanted get rid of the annoying warnings under the Github actions. See the blog post: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

**Changelog** of actions/checkout v3: https://github.com/actions/checkout/blob/main/CHANGELOG.md

> ## v3.0.0
>    
> Update to node 16 https://github.com/actions/checkout/pull/689

**Warnings** in Github Actions: https://github.com/simplepie/simplepie/actions/runs/4025417866

grafik

src/Idna/IdnaDomainFilter.php
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Convert an internationalized domain name.
+     *
+     * e.g. from `xn--mller-kva` to `müller`
+     *
+     * @param string $encoded The encoded domain name
+     *
+     * @return string the decoded domain name
+     */
+    public function filterDomain(string $encoded): string;
```
jtojnar, line 59: `filterDomain` does not really say anything about what is being done.

src/Enclosure.php
```diff
@@ -251,12 +249,8 @@ public function __construct($link = null, $type = null, $length = null, $javascr
         $this->type = $type;
         $this->width = $width;
 
-        if (class_exists('idna_convert')) {
-            $idn = new \idna_convert();
-            $parsed = \SimplePie\Misc::parse_url($link);
-            $this->link = \SimplePie\Misc::compress_parse_url($parsed['scheme'], $idn->encode($parsed['authority']), $parsed['path'], $parsed['query'], $parsed['fragment']);
```
jtojnar, line 257: The point of using `idna_convert` is that the user can type a domain name containing Unicode characters (so called [IDN](https://en.wikipedia.org/wiki/Internationalized_domain_name)) and it is encoded as ASCII (using [Punycode](https://en.wikipedia.org/wiki/Punycode) scheme), which various system APIs actually expect.

Though it looks like cURL, at least, can do this conversion too:

```php
php > $ch = curl_init('https://háčkyčárky.cz'); curl_setopt($ch, CURLOPT_RETURNTRANSFER, true); curl_exec($ch); var_dump(curl_error($ch));
string(0) ""
```

Unlike native PHP functions:

```php
php > var_dump(file_get_contents('https://háčkyčárky.cz'));

Warning: file_get_contents(): php_network_getaddresses: getaddrinfo for háčkyčárky.cz failed: Name or service not known in php shell code on line 1

Warning: file_get_contents(https://háčkyčárky.cz): Failed to open stream: php_network_getaddresses: getaddrinfo for háčkyčárky.cz failed: Name or service not known in php shell code on line 1
bool(false)
```

src/Idna/IdnaConverter.php
```diff
@@ -0,0 +1,87 @@
...
+        } else if (class_exists(IdnaConvert::class)) {
+            // Support for algo26-matthias/idna-convert:^2
+            $idnaConvert = new IdnaConvert();
+
+            return $idnaConvert->decode($encoded);
+        } else if (class_exists(idna_convert::class)) {
+            // Support for idna_convert:0.5.1
+            $idnaConvert = new idna_convert();
+
+            return $idnaConvert->decode($encoded);
```
jtojnar, line 81: But here you are switching it to decoding.

src/Item.php
```diff
@@ -2349,6 +2351,81 @@ public function get_source()
 
         return null;
     }
+
+    /**
+     * @uses idna_convert If available, this will convert an IDN in $url
+     * @see https://github.com/algo26-matthias/idna-convert
+     */
+    private function create_enclosure(
```
jtojnar, line 2359: Why move this out? It seems to me like it goes the opposite way from moving the IDN handling into the HTTP request component.

jtojnar (changes requested):

.github/workflows/ci.yml
Art4: I reverted the version change and opend #783

src/Item.php
Art4: I moved it here, because this is the only place where `Enclosure`s are created and `Item` already implements `RegistryAware`. If we want the idna_convert inside `Enclosure` we would have to implement `RegistryAware` there too.

src/Idna/IdnaDomainFilter.php
Art4: Naming things is hard. Any suggestions?

src/Enclosure.php
Art4: Oh, it seems that I mixed that up. 😅 Thank you, I will correct that.

src/Idna/IdnaConverter.php
Art4: ups... 😅

src/Enclosure.php
Art4: I fixed this by using the correct classes and methods to encode/convert the Unicode domain to ASCII.

src/Idna/IdnaConverter.php
Art4: Fixed, see https://github.com/simplepie/simplepie/pull/776#discussion_r1092947362

Art4:
Seems like that idna-convert@v0.5.1 has problems encoding some characters. 😕 

https://github.com/simplepie/simplepie/actions/runs/4062694656/jobs/6994049348#step:7:54

grafik

jtojnar:
Actually, maybe we could just use [`idn_to_ascii`](https://www.php.net/idn-to-ascii) function when present. And suggest https://github.com/symfony/polyfill-intl-idn in case people do not have `intl` extension.

Art4:
> Actually, maybe we could just use [`idn_to_ascii`](https://www.php.net/idn-to-ascii) function when present. And suggest https://github.com/symfony/polyfill-intl-idn in case people do not have `intl` extension.

I've tested native `idn_to_ascii()` but this function does something different then idn-convert (e.g. `ß` becomes `ss`). Thats why I added `IdnaDomainFilter`:

```php
use SimplePie\Idna\IdnaDomainFilter;

class MyIdnaConverter implements IdnaDomainFilter {
    public function filterDomain(string $decoded): string {
        return \idn_to_ascii($decoded);
    }
}

$simplepie = new \SimplePie\SimplePie();
$simplepie->get_registry()->register(IdnaDomainFilter::class, MyIdnaConverter::class);
```

Art4:
Maybe I should even remove support for Algo26\IdnaConvert v2 and v3 again in favor of `IdnaDomainFilter`. :thinking:

jtojnar:
PHP will default to old version of the standard. It needs to be passed `IDNA_NONTRANSITIONAL_TO_ASCII` (and `INTL_IDNA_VARIANT_UTS46` before PHP 7.4), see https://www.php.net/manual/en/function.idn-to-ascii.php#120917.

Art4:
> PHP will default to old version of the standard. It needs to be passed `IDNA_NONTRANSITIONAL_TO_ASCII` (and `INTL_IDNA_VARIANT_UTS46` before PHP 7.4), see https://www.php.net/manual/en/function.idn-to-ascii.php#120917.

Got it. `var_dump(idn_to_ascii('weißenbach', \IDNA_NONTRANSITIONAL_TO_ASCII, \INTL_IDNA_VARIANT_UTS46));` returns the expected `xn--weienbach-i1a`. https://3v4l.org/5HPJf

jtojnar:
I do not really think allowing choosing a custom IDN library is worth the extra piping. I would just keep it simple and replace the existing code and document the intl requirement:

```patch
diff --git a/README.markdown b/README.markdown
index 5500f03c..594d1322 100644
--- a/README.markdown
+++ b/README.markdown
@@ -14,6 +14,7 @@ Requirements
 * PHP 7.2+ (Required since SimplePie 1.8.0)
 * libxml2 (certain 2.7.x releases are too buggy for words, and will crash)
 * One of iconv, mbstring or intl extensions
+* intl extension or [symfony/polyfill-intl-idn](https://github.com/symfony/polyfill-intl-idn) to support IDNs
 * cURL or fsockopen()
 * PCRE support
 
@@ -29,10 +30,8 @@ What comes in the package?
    server for required settings.
 6. `demo/` - A basic feed reader demo that shows off some of SimplePie's more
    noticeable features.
-7. `idn/` - A third-party library that SimplePie can optionally use to
-   understand Internationalized Domain Names (IDNs).
-8. `build/` - Scripts related to generating pieces of SimplePie
-9. `test/` - SimplePie's unit test suite.
+7. `build/` - Scripts related to generating pieces of SimplePie
+8. `test/` - SimplePie's unit test suite.
 
 ### Where's `simplepie.inc`?
 Since SimplePie 1.3, we've split the classes into separate files to make it easier
diff --git a/demo/index.php b/demo/index.php
index 97cf2d15..47bc3178 100644
--- a/demo/index.php
+++ b/demo/index.php
@@ -6,7 +6,6 @@ $starttime = $starttime[1] + $starttime[0];
 // Include SimplePie
 // Located in the parent directory
 include_once('../autoloader.php');
-include_once('../idn/idna_convert.class.php');
 
 // Create a new instance of the SimplePie object
 $feed = new \SimplePie\SimplePie();
diff --git a/demo/test.php b/demo/test.php
index 47bed901..18764106 100644
--- a/demo/test.php
+++ b/demo/test.php
@@ -1,6 +1,5 @@
 <?php
 include_once('../autoloader.php');
-include_once('../idn/idna_convert.class.php');
 
 // Parse it
 $feed = new \SimplePie\SimplePie();
diff --git a/src/Enclosure.php b/src/Enclosure.php
index 124d65b6..26e9133c 100644
--- a/src/Enclosure.php
+++ b/src/Enclosure.php
@@ -221,7 +221,7 @@ class Enclosure
      * For documentation on all the parameters, see the corresponding
      * properties and their accessors
      *
-     * @uses idna_convert If available, this will convert an IDN
+     * @uses intl extension If available, this will convert an IDN
      */
     public function __construct($link = null, $type = null, $length = null, $javascript = null, $bitrate = null, $captions = null, $categories = null, $channels = null, $copyright = null, $credits = null, $description = null, $duration = null, $expression = null, $framerate = null, $hashes = null, $height = null, $keywords = null, $lang = null, $medium = null, $player = null, $ratings = null, $restrictions = null, $samplingrate = null, $thumbnails = null, $title = null, $width = null)
     {
@@ -251,10 +251,10 @@ class Enclosure
         $this->type = $type;
         $this->width = $width;
 
-        if (class_exists('idna_convert')) {
-            $idn = new \idna_convert();
+        if (function_exists('idn_to_ascii')) {
             $parsed = \SimplePie\Misc::parse_url($link);
-            $this->link = \SimplePie\Misc::compress_parse_url($parsed['scheme'], $idn->encode($parsed['authority']), $parsed['path'], $parsed['query'], $parsed['fragment']);
+            $authority = \idn_to_ascii($parsed['authority'], IDNA_NONTRANSITIONAL_TO_ASCII, INTL_IDNA_VARIANT_UTS46);
+            $this->link = \SimplePie\Misc::compress_parse_url($parsed['scheme'], $authority, $parsed['path'], $parsed['query'], $parsed['fragment']);
         }
         $this->handler = $this->get_handler(); // Needs to load last
     }
diff --git a/src/File.php b/src/File.php
index bc18b8d8..2a92f1b2 100644
--- a/src/File.php
+++ b/src/File.php
@@ -68,10 +68,10 @@ class File
 
     public function __construct($url, $timeout = 10, $redirects = 5, $headers = null, $useragent = null, $force_fsockopen = false, $curl_options = [])
     {
-        if (class_exists('idna_convert')) {
-            $idn = new \idna_convert();
+        if (function_exists('idn_to_ascii')) {
             $parsed = \SimplePie\Misc::parse_url($url);
-            $url = \SimplePie\Misc::compress_parse_url($parsed['scheme'], $idn->encode($parsed['authority']), $parsed['path'], $parsed['query'], null);
+            $authority = \idn_to_ascii($parsed['authority'], IDNA_NONTRANSITIONAL_TO_ASCII, INTL_IDNA_VARIANT_UTS46);
+            $url = \SimplePie\Misc::compress_parse_url($parsed['scheme'], $authority, $parsed['path'], $parsed['query'], null);
         }
         $this->url = $url;
         $this->permanent_url = $url;
```

Art4:
> I do not really think allowing choosing a custom IDN library is worth the extra piping. I would just keep it simple and replace the existing code and document the intl requirement:

I like the idea. :+1: This would make everything easier. Because of `functions_exists()` the dependency to IDNA support will be optional, and we can remove the buggy 0.5.1 class. But I think we will have to modify some tests.

Art4:
@jtojnar Would you create a PR for your proposal? Then I could close this PR in favor of yours.

PR #777, opened 2023-01-27
Add support for PSR-18 HTTP clients

### LINKED ISSUE #520: Migration to a HTTP library

Art4:
Hi all 👋 

This is a follow up of #774 and will add support for every [PSR-18](https://www.php-fig.org/psr/psr-18/) HTTP client implementation. 🥳 

This PR will fixes #520. I propose this PR for SimplePie 1.9.0, see #731.


src/Content/Type/Sniffer.php
```diff
@@ -71,8 +73,17 @@ class Sniffer
      *
      * @param File $file Input file
      */
-    public function __construct(File $file)
+    public function __construct(/* File */ $file)
```
Art4, line 76: Removing the type hint is no breaking change because we have not released 1.9.0

Art4:
This PR is ready for review. Because this PR is build on #774 I recommend to merge this PR first. If will fix some BC breaks that comes with #774.

To show only the changes between #774 and #777 you can look at the diff between the branches: https://github.com/Art4/simplepie/compare/split-file-into-client-and-response...Art4:simplepie:add-psr18-http-client-support

src/HTTP/Psr18Client.php
```diff
@@ -0,0 +1,144 @@
...
+                    break;
+                }
+
+                $remainingRedirects--;
+                $followRedirect = true;
+
+                $requestedUrl = $response->getHeaderLine('Location');
+
+                if ($statusCode === 301) {
+                    $permanentUrl = $requestedUrl;
```
jtojnar, line 120: We need some kind of mechanism to not change the permanent URL after temporary redirects (as tested by https://github.com/simplepie/simplepie/pull/819). For example, this is what File does:

https://github.com/simplepie/simplepie/blob/e356f5dc711ad4c5d92091a7dea509bf7c1c11a6/src/File.php#L144-L144

jtojnar (changes requested):
Still need to resolve the question of [non-absolute Location header](https://github.com/simplepie/simplepie/pull/826#discussion_r1142088339) and [Redirect tracking](https://github.com/simplepie/simplepie/pull/826#discussion_r1227293061).

mblaney:
thanks @Art4 this looks good, can we merge and handle the redirect case separately?

Art4:
@mblaney I resolved all conflicts and fixed all PHPStan errors.

PR #779, opened 2023-01-28
Typehinting

ColonelMoutarde:


jtojnar:
Adding return typehints would break subclasses. And param typehints are covered by #775

Issue #780, opened 2023-01-30
[ENH] Revise the phpdoc tags `@package` and `@subpackage`


Art4:
####  Is your feature request related to a problem? Please describe.

To quote @jtojnar in [this comment](https://github.com/simplepie/simplepie/pull/774#discussion_r1089710022):

> [`subpackage`](https://docs.phpdoc.org/2.9/references/phpdoc/tags/subpackage.html) is deprecated and [`package`](https://docs.phpdoc.org/2.9/references/phpdoc/tags/package.html) is redundant.

#### Describe the solution you'd like

Quote from @jtojnar in [this comment](https://github.com/simplepie/simplepie/pull/774#discussion_r1089777691):

> The [`subpackage`](https://docs.phpdoc.org/2.9/references/phpdoc/tags/subpackage.html) documentation suggests `@package SimplePie\HTTP`, and [`package`](https://docs.phpdoc.org/2.9/references/phpdoc/tags/package.html) docs say it is not necessary when it is the same as the namespace. I would just drop both.

Issue #778, opened 2023-01-28
Use SPDX comments


jtojnar:
Currently, each file starts with a long license blurb. It might be cleaner to replace it with SPDX comments, which also uniquely identify the license, while being much shorter:

```php
// SPDX-License-Identifier: BSD-3-Clause
// SPDX-FileCopyrightText: 2022 Artur Weigandt
```

- https://reuse.software/tutorial/
- https://www.linuxfoundation.org/blog/blog/solving-license-compliance-at-the-source-adding-spdx-license-ids

_Originally posted by @jtojnar in https://github.com/simplepie/simplepie/pull/752#discussion_r1005729998_

Art4:
I like the concept. 👍 

It took me a while to find a PHP repository that has implement the REUSE spec correctly. (most repos don't have the requires `LICENSES` folder)

Here one can see a good example how the licenses will be managed: https://github.com/theCalcaholic/nextcloud-secrets

For me it looks light and clean. And I like that even images or binary files will have a specific license information.

PR #781, opened 2023-01-30
Remove package phpdoc annotation

jtojnar:
`@subpackage` is deprecated and `@package` is mostly redundant with PHP namespaces.

And while the annotations sometimes do not match the namespaces perfectly, which could be taken as a separate logical structure, having two separate hierarchies is IMO more confusing than helpful. Especially when the annotations are not even picked up by ApiGen.

https://docs.phpdoc.org/2.9/references/phpdoc/tags/package.html
https://docs.phpdoc.org/2.9/references/phpdoc/tags/subpackage.html
https://github.com/ApiGen/ApiGen/issues/1081

Removed with the following commands:

    sd '(?m)(?:\n^[ \t]+\*)?\n[ \t]+\*[ \t]+@package.+?(?:\n[ \t]+\*[ \t]+@subpackage.+)?(\n[ \t]+\*/|(\n[ \t]*\*))' '$2$1' "$(rg -l @subpackage -t php)"

Fixes: https://github.com/simplepie/simplepie/issues/780

### DIFF
--- a/autoloader.php
+++ b/autoloader.php
@@ -32,7 +32,6 @@
  * OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  *
- * @package SimplePie
  * @copyright 2004-2016 Ryan Parman, Sam Sneddon, Ryan McCue
  * @author Ryan Parman
  * @author Sam Sneddon
@@ -93,9 +92,6 @@ if (!class_exists('SimplePie'))
 
 /**
  * Autoloader class
- *
- * @package SimplePie
- * @subpackage API
  */
 class SimplePie_Autoloader
 {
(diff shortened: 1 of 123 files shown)


Art4:
If ApiGen shows the namespaces, then the "packages" are not needed. I have 3 notes/questions:

1. there is another `@package` in `build/header.txt` that could be removed.
2. While we are on the phpdoc topic I wonder how the `{@see ...}` syntax works. E.g. see https://github.com/simplepie/simplepie/blob/1.8.0/src/SimplePie.php#L1411-L1432

    But ApiGen seems to ignores the syntax and shows it in plain text: http://simplepie.ogion.cz/api/SimplePie.SimplePie/#_set_output_encoding
  
    How should this normally works? Can we fix this in this PR?
3. Can we fix the copyright year in all files? There are places where the year is `2004-2016` or `2004-2022`.

jtojnar:
> 2. While we are on the phpdoc topic I wonder how the `{@see ...}` syntax works. E.g. see [`1.8.0`/src/SimplePie.php#L1411-L1432](https://github.com/simplepie/simplepie/blob/1.8.0/src/SimplePie.php?rgh-link-date=2023-01-30T19%3A38%3A08Z#L1411-L1432)

It worked in older ApiGen versions, newer ones expect Markdown in PHPDoc descriptions.

> 3. Can we fix the copyright year in all files? There are places where the year is `2004-2016` or `2004-2022`.

I would rather handle that with #778.

> 1. there is another `@package` in `build/header.txt` that could be removed.

Removed.

Art4 (approved):

PR #784, opened 2023-02-08
Use single constant for default HTTP Accept header

Alkarex:
The same string was repeated 5 times.
And add missing headers in `SimplePie_Locator::body()`
Follow-up of https://github.com/simplepie/simplepie/commit/5d966b9f64a6034f89b019152d7cad9059f8819f
This prepares for an update of the content of the default HTTP Accept header in another PR.

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -55,6 +55,9 @@ namespace SimplePie;
  */
 class File
 {
+
+    const DEFAULT_HTTP_ACCEPT = 'application/atom+xml, application/rss+xml, application/rdf+xml;q=0.9, application/xml;q=0.8, text/xml;q=0.8, text/html;q=0.7, unknown/unknown;q=0.1, application/unknown;q=0.1, */*;q=0.1';
+
     public $url;
     public $useragent;
     public $success = true;
--- a/src/Locator.php
+++ b/src/Locator.php
@@ -221,7 +221,7 @@ class Locator implements RegistryAware
                 if (!in_array($href, $done) && in_array('feed', $rel) || (in_array('alternate', $rel) && !in_array('stylesheet', $rel) && $link->hasAttribute('type') && in_array(strtolower($this->registry->call(Misc::class, 'parse_mime', [$link->getAttribute('type')])), ['text/html', 'application/rss+xml', 'application/atom+xml'])) && !isset($feeds[$href])) {
                     $this->checked_feeds++;
                     $headers = [
-                        'Accept' => 'application/atom+xml, application/rss+xml, application/rdf+xml;q=0.9, application/xml;q=0.8, text/xml;q=0.8, text/html;q=0.7, unknown/unknown;q=0.1, application/unknown;q=0.1, */*;q=0.1',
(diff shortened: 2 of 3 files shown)


src/Locator.php
```diff
@@ -350,9 +350,9 @@ public function body(&$array)
             if (preg_match('/(feed|rss|rdf|atom|xml)/i', $value)) {
                 $this->checked_feeds++;
                 $headers = [
-                    'Accept' => 'application/atom+xml, application/rss+xml, application/rdf+xml;q=0.9, application/xml;q=0.8, text/xml;q=0.8, text/html;q=0.7, unknown/unknown;q=0.1, application/unknown;q=0.1, */*;q=0.1',
+                    'Accept' => File::DEFAULT_HTTP_ACCEPT,
                 ];
-                $feed = $this->registry->create(File::class, [$value, $this->timeout, 5, null, $this->useragent, $this->force_fsockopen, $this->curl_options]);
+                $feed = $this->registry->create(File::class, [$value, $this->timeout, 5, $headers, $this->useragent, $this->force_fsockopen, $this->curl_options]);
```
Alkarex, line 355: Likely forgotten in https://github.com/simplepie/simplepie/commit/5d966b9f64a6034f89b019152d7cad9059f8819f
Otherwise, the `$headers` was not used.

Alkarex:
Downstream PR https://github.com/FreshRSS/FreshRSS/pull/5083

src/File.php
```diff
@@ -55,6 +55,9 @@
  */
 class File
 {
+
+    const DEFAULT_HTTP_ACCEPT = 'application/atom+xml, application/rss+xml, application/rdf+xml;q=0.9, application/xml;q=0.8, text/xml;q=0.8, text/html;q=0.7, unknown/unknown;q=0.1, application/unknown;q=0.1, */*;q=0.1';
```
Art4, line 59: I suggest that we define this constant as `DEFAULT_HTTP_ACCEPT_HEADER` in the `SimplePie` class and mark it as `@internal`, since the `File` class will soon be deprecated.

src/Locator.php
Art4: 👍 
Looks like an error to me. I am in favor of fixing it this way.

Art4 (changes requested):

src/File.php
Alkarex: Done

Art4 (approved):
Thank you 👍

src/SimplePie.php
```diff
@@ -407,6 +407,11 @@ class SimplePie
      */
     public const FILE_SOURCE_FILE_GET_CONTENTS = 16;
 
+    /**
+     * @internal Default HTTP Accept header when sending HTTP requests
```
jtojnar, line 411: Maybe something like:

```suggestion
     * @internal Default value of the HTTP Accept header when fetching/locating feeds
```

We probably will not want it for e.g. downloading and caching images. And header would consist of the `Accept: ` prefix so mention it is a value.

PR #785, opened 2023-02-16
Use `idn_to_ascii` function instead of `idna_convert` library

jtojnar:
The `idna_convert` library is outdated and there is now a native PHP function to convert IDNs to punycode.
It requires `intl` extension or a [polyfill](https://github.com/symfony/polyfill-intl-idn).

Supersedes #776


src/Enclosure.php
```diff
@@ -251,10 +251,12 @@ public function __construct($link = null, $type = null, $length = null, $javascr
         $this->type = $type;
         $this->width = $width;
 
-        if (class_exists('idna_convert')) {
-            $idn = new \idna_convert();
+        if (function_exists('idn_to_ascii')) {
             $parsed = \SimplePie\Misc::parse_url($link);
-            $this->link = \SimplePie\Misc::compress_parse_url($parsed['scheme'], $idn->encode($parsed['authority']), $parsed['path'], $parsed['query'], $parsed['fragment']);
+            if ($parsed['authority'] !== '' && !ctype_print($parsed['authority'])) {
+                $authority = \idn_to_ascii($parsed['authority'], \IDNA_NONTRANSITIONAL_TO_ASCII, \INTL_IDNA_VARIANT_UTS46);
```
jtojnar, line 257: It looks like the original code was already broken on authority strings containing non-ASCII characters in userinfo part of the URI, e.g. `čížek@háčkyčárky.cz`:

- This is what `idna_convert` gives: `xn--ek-mja1kox@xn--hkyrky-ptac70bc.cz`
- This is what Firefox converts: `%C4%8D%C3%AD%C5%BEek@xn--hkyrky-ptac70bc.cz`

It probably does not make sense to attempt to fix it when it will likely be corrected with switch to PSR-18.

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - PHP 7.2 or newer is now required by @Art4 in [#743](https://github.com/simplepie/simplepie/pull/743)
 - Parse\Date: Clean up regex structure by @jtojnar in [#765](https://github.com/simplepie/simplepie/pull/765)
 - Declare strict_types=1 in every file by @Art4 in [#763](https://github.com/simplepie/simplepie/pull/763)
+- Use `idn_to_ascii` function instead of `idna_convert` library (requires `intl` extension or a [polyfill](https://github.com/symfony/polyfill-intl-idn)) by @jtojnar in [#785](https://github.com/simplepie/simplepie/pull/785)
```
Art4, line 28: This should moved up to the `Unreleased` section.

README.markdown
```diff
@@ -14,6 +14,7 @@ Requirements
 * PHP 7.2+ (Required since SimplePie 1.8.0)
 * libxml2 (certain 2.7.x releases are too buggy for words, and will crash)
 * One of iconv, mbstring or intl extensions
+* intl extension or [symfony/polyfill-intl-idn](https://github.com/symfony/polyfill-intl-idn) to support IDNs
```
Art4, line 17: This is not really a requirement, but an optional feature. Maybe we should create a new "optional feature" section?

idn/idna_convert.class.php
```diff
@@ -1,969 +0,0 @@
-<?php
-// {{{ license
```
Art4, line 2: Removing this file might be a breaking change. What do you think about keeping this file and simply triggering a deprecation error?

```php
<?php

declare(strict_types=1);

trigger_error('Using class `idna_convert` from `idn/idna_convert.class.php` is deprecated since SimplePie 1.9.0, use native `idn_to_ascii()` function instead.', \E_USER_DEPRECATED);
```

Art4 (changes requested):

Art4:
I like this approach. `idn_to_ascii()` should available in the most PHP environments and will work out of the box without requiring any files. If it is not present, one can define it using the `Algo26\IdnaConvert` library or using the polyfill `symfony/polyfill-intl-idn`.

This PR fixes #538.

idn/idna_convert.class.php
jtojnar: I would argue that this is not a part of API – it is not part of SimplePie, it is not listed in the API docs nor was it documented anywhere else.

I have listed it in the Removed section of changelog so that if someone uses it for themselves, they can just copy it from older version.

Art4 (approved):

Art4:
ping @mblaney

PR #789, opened 2023-02-18
Misc: Fix get_build when overridden

jtojnar:
Introduced in https://github.com/simplepie/simplepie/commit/a2a22687ecebe8f13a803410936e67a43f4cc939, `Misc::SIMPLEPIE_BUILD` is private so accessing it through `static` will not work unless the child class redefines the property.

Needed for PHPStan level 2 to pass.

### DIFF
--- a/src/Misc.php
+++ b/src/Misc.php
@@ -2074,15 +2074,15 @@ END;
      */
     public static function get_build()
     {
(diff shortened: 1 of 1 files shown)


Art4 (approved):
Nice catch :+1:

PR #791, opened 2023-02-18
IRI: Fix types

jtojnar:
Needed for PHPStan level 2 to pass.

This is not complete (there are more magic properties) and the types are conservative (e.g. path accepts setting `null` which will be converted to an empty string) but this class is going away in the future and this is sufficient for our internal use.


Art4 (approved):

jtojnar:
Added `userinfo`, `host` and `port` since those are also used by tests.

PR #792, opened 2023-02-18
Do not call pure Misc methods through registry

jtojnar:
These have a singular purpose and except for parse_uri do not interact with other classes. Overriding them makes little sense.

This makes the code a bit cleaner and less opaque for static analysis.

I am ignoring Misc::error since it will go away.

Partly reverts https://github.com/simplepie/simplepie/commit/f6e63e6ba8e4a198283bb2323aa498fd0d6a216e


Art4 (changes requested):
As much as I like this PR, but this would be a breaking change. Instead we could create non-static methods in Misc class and calling them with `$registry->create(Misc::class)->...`. Then we can deprecate the `$registry->call()` method.

jtojnar:
I do not see what does this change break. It changes the behaviour that the methods will no longer be overridable but I do not see how it would break any promises, any more than any other implementation change would.

Art4:
Something like this will not work anymore:

```php
<?php

class CustomMisc extends \SimplePie\Misc
{
    public static function fix_protocol($url, $http = 1)
    {
        // custom implementation ...
        return $url;
    }
}

$simplepie = new \SimplePie\SimplePie();
$simplepie->get_registry()->register('Misc', CustomMisc::class);

$simplepie->set_feed_url('url_where_the_protocol_should_be_fixed_with_my_implementation');
```

PR #798, opened 2023-02-19
Replace the licence blurbs with SPDX comments

jtojnar:
SPDX is a standard for describing licensing and copyright status of project files that has been gaining traction lately.
In addition to not filling more than half a screen, it is machine readable and can be used to produce Software Bill of Materials.
We are keeping the long header in the compiled single-file bundle since that might be copied into third-party projects in a way that makes it easy to accidentally forget also copying the license text.

https://reuse.software/tutorial/
https://www.linuxfoundation.org/blog/blog/solving-license-compliance-at-the-source-adding-spdx-license-ids

Fixes: https://github.com/simplepie/simplepie/issues/778


Art4:
Is this PR ready for review? Because I'm missing the `LICENSES` folder with the licenses.

jtojnar:
Currently, there is only a single license so I kept it in the root.

Art4:
So, why not move the `LICENSE.txt` to `LICENSES/BSD-3-Clause.txt` and replace the content with the official one?

https://github.com/spdx/license-list-data/blob/v3.20/text/BSD-3-Clause.txt

jtojnar:
Done.

PR #799, opened 2023-02-21
Fixed PHP 8.1: strftime and gmstrftime functions are deprecated

raortegar:
#### When submitting a pull request

1. Run the linter.

   ```bash
   make lint
   ```

1. Run the tests.

   ```bash
   make test
   ```

1. Carefully _read your diff_ before submitting a PR.
    * Make sure your diff clearly represents your changes.
    * Make sure your code is easy for reviewers to follow.
    * Can your code review be broken into smaller chunks?
    * Make sure you have PHPUnit tests that cover your changes.

1. Make sure that your changes are very-nearly complete.
    * Small changes/fixes based on feedback are OK.
    * Major changes should be avoided while a PR is open. If you need to make major changes, close the PR, make your changes, and open a new PR once you're ready.

1. In the description, start with an explanation of the change and why you want to make it.
    * Make any relevant documentation easily available for reviewers.

1. Add a list of changes you've made using [task list](https://help.github.com/articles/basic-writing-and-formatting-syntax/#task-lists) syntax.
    * Things that are complete should have `[X]`.
    * Things that are still incomplete should have `[ ]`.

1. Discuss feedback.
    * The _SimplePie NG Core Team_ has the final say on what gets accepted.
    * Respond to all code review feedback.
    * _Respectfully_ discuss any feedback you disagree with. You may need to bring a _better_ argument to the table to convince us.

1. If you don't understand something, _ask_.

### DIFF
--- a/src/Item.php
+++ b/src/Item.php
@@ -746,7 +746,7 @@ class Item implements RegistryAware
         if (!$date_format) {
             return $this->sanitize($this->get_date(''), \SimplePie\SimplePie::CONSTRUCT_TEXT);
         } elseif (($date = $this->get_date('U')) !== null && $date !== false) {
-            return strftime($date_format, $date);
+            return date($date_format, $date);
         }
 
         return null;


jtojnar (changes requested):
Thanks. Unfortunately, the `date` function accepts a different format, even the default value of the `$date_format` argument is not supported (compare https://www.php.net/manual/en/datetime.format.php#refsect1-datetime.format-parameters and https://www.php.net/manual/en/function.strftime.php#refsect1-datetime.format-parameters).

I would suggest deprecating `get_local_date` method and adding `get_date_string(): string` and `get_datetime(): ?DateTime` methods, corresponding to the first and second `if` branches, respectively. The local formatting would then be a responsibility of the consumer.

Issue #802, opened 2023-03-17
[BUG] must be of type string, bool given in /framework/simplepie/src/Locator.php:330


thE-iNviNciblE:
> **IMPORTANT:** If you are posting code or output, please make sure you wrap it with [fenced code blocks](https://help.github.com/articles/creating-and-highlighting-code-blocks/) so that it is readable.

#### Description

PHP Fatal error:  Uncaught TypeError: strtolower(): Argument #1 ($string) must be of type string, bool given in /var/www/vhosts/xxx/framework/simplepie/src/Locator.php:330

` if (in_array(strtolower(strrchr($value, '.')), ['.rss', '.rdf', '.atom', '.xml'])) {`

edit to 

```
            $str = strrchr($value, '.');
            if(is_string($str)) {

                if (in_array(strtolower($str), ['.rss', '.rdf', '.atom', '.xml'])) {
```

#### Steps to Reproduce

it's a cron.php which is running, can't excatly say which is the $value but 
#### Actual Behavior

`strrchr(string $haystack, string $needle): string|false`

it's returning "false" in case of an error not a string and strtolower($str) wants that is a string.

#### Possible Solutions
```

 $str = strrchr($value, '.');
            if(is_string($str)) {

                if (in_array(strtolower($str), ['.rss', '.rdf', '.atom', '.xml'])) {
                    $this->checked_feeds++;
    
                    $headers = [
                        'Accept' => 'application/atom+xml, application/rss+xml, application/rdf+xml;q=0.9, application/xml;q=0.8, text/xml;q=0.8, text/html;q=0.7, unknown/unknown;q=0.1, application/unknown;q=0.1, */*;q=0.1',
                    ];
                    $feed = $this->registry->create(File::class, [$value, $this->timeout, 5, $headers, $this->useragent, $this->force_fsockopen, $this->curl_options]);
                    if ($feed->success && ($feed->method & \SimplePie\SimplePie::FILE_SOURCE_REMOTE === 0 || ($feed->status_code === 200 || $feed->status_code > 206 && $feed->status_code < 300)) && $this->is_feed($feed)) {
                        return [$feed];
                    } else {
                        unset($array[$key]);
                    }
                }
            }
```

#### Environment

Normal System Debian 10 with PHP 8.2

jtojnar:
Thanks for reporting, opened https://github.com/simplepie/simplepie/pull/803

thE-iNviNciblE:
no problem.

[...]

PR #803, opened 2023-03-17
Locator: Fix type error when no extension

jtojnar:
Fixes https://github.com/simplepie/simplepie/issues/802


Art4 (approved):
A test would be nice, but LGTM. 👍

jtojnar:
#819 will allow us to do integration tests for locator.

Issue #810, opened 2023-03-18
[BUG] PHP Fatal error:  Uncaught TypeError: strspn(): Argument #1 ($string) must be of type string, null given in /var/www/vhosts/xxx/framework/simplepie/src/Content/Type/Sniffer.php:153


thE-iNviNciblE:
> **IMPORTANT:** If you are posting code or output, please make sure you wrap it with [fenced code blocks](https://help.github.com/articles/creating-and-highlighting-code-blocks/) so that it is readable.

#### Description

`$ws = strspn($this->file->body, "\x09\x0A\x0B\x0C\x0D\x20");` - simplepie/src/Content/Type/Sniffer.php:153

$this->file->body is empty.

#### Steps to Reproduce

it's maybe a empty resource where the DOM HTML is empty.

#### Expected Behavior

skip invalid RSS-Source.

#### Actual Behavior

Fatal error cron job exits.

#### Possible Solutions

strlen() Check should work to avoid strspn() to crash.
```
        if(strlen($this->file->body) != 0) {
        
            $ws = strspn($this->file->body, "\x09\x0A\x0B\x0C\x0D\x20");
            if (strtolower(substr($this->file->body, $ws, 14)) === '<!doctype html'
                || strtolower(substr($this->file->body, $ws, 5)) === '<html'
                || strtolower(substr($this->file->body, $ws, 7)) === '<script') {
                return 'text/html';
            } elseif (substr($this->file->body, 0, 5) === '%PDF-') {
                return 'application/pdf';
            } elseif (substr($this->file->body, 0, 11) === '%!PS-Adobe-') {
                return 'application/postscript';
            } elseif (substr($this->file->body, 0, 6) === 'GIF87a'
                || substr($this->file->body, 0, 6) === 'GIF89a') {
                return 'image/gif';
            } elseif (substr($this->file->body, 0, 8) === "\x89\x50\x4E\x47\x0D\x0A\x1A\x0A") {
                return 'image/png';
            } elseif (substr($this->file->body, 0, 3) === "\xFF\xD8\xFF") {
                return 'image/jpeg';
            } elseif (substr($this->file->body, 0, 2) === "\x42\x4D") {
                return 'image/bmp';
            } elseif (substr($this->file->body, 0, 4) === "\x00\x00\x01\x00") {
                return 'image/vnd.microsoft.icon';
            }
            
            return $this->text_or_binary();
        }
```

#### Environment

Debian 10 with PHP 8.2

[...]

PR #811, opened 2023-03-18
ci: Add PHPStan

jtojnar:
Ideally, this would prevent issues like https://github.com/simplepie/simplepie/issues/802, https://github.com/simplepie/simplepie/issues/810, …

But this is just a beginning, targetting level 2 for now.

Depends on:

- #793 
- #786 
- #791 
- #790 
- #787 
- #789 
- #788 
- #794 
- #816
- #817
- #820
- #823
- #824

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,9 @@ jobs:
       - name: Run unit tests
         run: composer test
 
+      - name: Run static analysis
+        run: composer phpstan -- --error-format=github
+
   test-compiled:
     runs-on: ubuntu-latest
 
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,7 @@
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.19 || ^3.8",
         "mf2/mf2": "^0.5.0",
+        "phpstan/phpstan": "^1.10",
         "psr/simple-cache": "^1 || ^2 || ^3",
         "yoast/phpunit-polyfills": "^1.0.1"
     },
@@ -63,6 +64,7 @@
     "scripts": {
         "cs": "php-cs-fixer fix --verbose --dry-run --diff",
         "fix": "php-cs-fixer fix --verbose --diff",
+        "phpstan": "phpstan analyze --memory-limit 512M",
         "test": "phpunit"
     }
 }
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -46,6 +46,9 @@ jobs:
       - name: Run unit tests
         run: composer test
 
+      - name: Run static analysis
+        run: composer phpstan
```
Art4, line 50: What do you think about running PHPStan before running composer test?

jtojnar: Since we use `yoast/phpunit-polyfills`, PHPUnit will not be available until tests are run.

.github/workflows/ci.yml
```diff
@@ -30,6 +30,8 @@ jobs:
           ini-values: error_reporting=-1, display_errors=On, log_errors_max_len=0
           coverage: none
           tools: none
+          # Not enabled on PHP ≥ 8.2 by default
+          extensions: xmlreader
```
jtojnar, line 34: Ugh, PHPStan complains even with this on PHP ≥ 8.2 for some reason:

```
Error: Access to an undefined property XMLReader::$nodeType.
Error: Access to an undefined property XMLReader::$namespaceURI.
Error: Access to an undefined property XMLReader::$localName.
Error: Access to an undefined property XMLReader::$isEmptyElement.
Error: Access to an undefined property XMLReader::$value.
```

jtojnar: Can reproduce locally, opened https://github.com/phpstan/phpstan/issues/9057.

src/IRI.php
```diff
@@ -10,34 +10,43 @@
 
 /**
  * IRI parser/serialiser/normaliser
+ *
+ * @property ?string $scheme
+ * @property ?string $userinfo
+ * @property ?string $host
+ * @property ?int $port
```
jtojnar, line 17: For level 3, we would need something like:

```suggestion
 * @property-read int|null $port
 * @property-write int|string|null $port
```

But that is not currently supported https://github.com/phpstan/phpstan/issues/9062

jtojnar:
Dropped https://github.com/simplepie/simplepie/pull/792 and https://github.com/simplepie/simplepie/pull/785 from dependencies since they are not necessary for level 2 but they will be needed for higher levels.

Art4 (approved):

PR #812, opened 2023-03-18
Fix `permanent_url` with multiple permanent redirects

jtojnar:
The `File` class handles redirects by calling the constructor recursively. To avoid setting the `permanent_url` incorrectly at the top of the constructor, the value of the property was reset after each constructor call. And since local variables were used, only the reset following the first internal constructor call would actually do anything. This was very confusing and would stop considering permanent redirects after the first one.
Let’s avoid doing stuff after popping the stack, tail recursion FTW.

Additionally, the `permanent_url` would only be updated when `HTTP 301 Moved Permanently` status code was encountered.
Let’s also support `HTTP 308 Permanent Redirect`.

Tracking the mutability in a property is not the cleanest but a bigger refactoring is probably not worth it with the upcoming deprecation of File.

Unfortunately, we cannot easily test this since we cannot mock curl.

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -55,6 +55,7 @@ namespace SimplePie;
  */
 class File
 {
+    /** @var string The final URL after following all redirects */
     public $url;
     public $useragent;
     public $success = true;
@@ -64,7 +65,10 @@ class File
     public $redirects = 0;
     public $error;
     public $method = \SimplePie\SimplePie::FILE_SOURCE_NONE;
+    /** @var string The permanent URL or the resource (first URL after the prefix of (only) permanent redirects) */
     public $permanent_url;
+    /** @var bool Whether the permanent URL is still writeable (prefix of permanent redirects has not ended) */
+    private $permanentUrlMutable = true;
 
     public function __construct($url, $timeout = 10, $redirects = 5, $headers = null, $useragent = null, $force_fsockopen = false, $curl_options = [])
     {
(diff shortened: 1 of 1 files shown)


jtojnar:
I managed to test this with

```patch
--- a/tests/Unit/FileTest.php
+++ b/tests/Unit/FileTest.php
@@ -45,6 +45,7 @@ declare(strict_types=1);
 namespace SimplePie\Tests\Unit;
 
 use PHPUnit\Framework\TestCase;
+use SimplePie\File;
 
 class FileTest extends TestCase
 {
@@ -57,4 +58,13 @@ class FileTest extends TestCase
     {
         $this->assertTrue(class_exists('SimplePie_File'));
     }
+
+    public function testFoo() {
+        $file = new File(
+            url: 'http://localhost:8000/perm2',
+            redirects: 10,
+        );
+        $this->assertSame('http://localhost:8000/temp2', $file->permanent_url);
+        $this->assertSame('http://localhost:8000/final', $file->url);
+    }
 }
```

while running the following script with `php -S localhost:8000 srv.php`:

```php
<?php

function redirect(int $status, string $location): void {
    http_response_code($status);
    header("Location: $location");
}

function output(int $status, string $text): void {
    http_response_code($status);
    echo $text;
}

match ($_SERVER['REQUEST_URI']) {
    '/perm2' => redirect(308, '/perm1'),
    '/perm1' => redirect(301, '/perm0'),
    '/perm0' => redirect(308, '/temp2'),
    '/temp2' => redirect(307, '/temp1'),
    '/temp1' => redirect(302, '/temp0'),
    '/temp0' => redirect(307, '/permA'),
    '/permA' => redirect(301, '/permB'),
    '/permB' => redirect(308, '/permC'),
    '/permC' => redirect(308, '/final'),
    '/final' => output(200, '/ok'),
    default => output(404, 'Not found'),
};
```

both with `curl` and `fsockopen`.

src/File.php
jtojnar, line 1: Primary motivation is to make `Response::get_permanent_uri` from #774 meaningful.

PR #815, opened 2023-03-19
Split SimplePie\File class into HTTP client and response

jtojnar:
This is https://github.com/simplepie/simplepie/pull/774, with few clean-ups for easier reviewing commit-by-commit (commits squashed or split to ideally fit into my head, added some implementation notes into commit messages).

The main differences as outlined in https://github.com/simplepie/simplepie/pull/774#pullrequestreview-1347330706:
- Renamed  `Response::get_requested_uri()` to `Response::get_final_requested_uri()` so it is clearer what this does.
- We do not expect `Response::get_headers()` to preserve header capitalization, removed mention of that from the comment.
- Using SPDX to credit PHP Framework Interoperability Group, as the documentation comments taken from PSR-7 are non-trivial.
- Adding support for `HTTP\Parser` to produce PSR-7 compatible headers, using `$parsed_headers` internally in `File` as $774 did initially, and recomputing `$parsed_headers` when `$headers` change (which should not really happen outside of tests but better safe than sorry).
- Extracting the is_remote check to `Misc::is_remote_uri()`.
- Passing the reduced timeout to `SimplePie::get_http_client()` factory instead of side-channel signalling through `$this->timeout`.
- Fixing early loop terminations in `Locator`.

[Full diff](https://github.com/Art4/simplepie/compare/split-file-into-client-and-response..jtojnar:simplepie:split-file-into-client-and-response) (will be cleaner once dependencies are merged).

Depends on:
- #798
- https://github.com/simplepie/simplepie/pull/808
- https://github.com/simplepie/simplepie/pull/805

Should be mergable without squashing (e.g. rebase).


src/SimplePie.php
```diff
@@ -1742,25 +1745,27 @@ protected function fetch_data(&$cache)
...
-                            if ($file->status_code === 304) {
-                                // Set raw_data to false here too, to signify that the cache
-                                // is still valid.
-                                $this->raw_data = false;
-                                $cache->set_data($cacheKey, $this->data, $this->cache_duration);
-                                return true;
-                            }
-                        } else {
+                        try {
+                            $file = $this->get_http_client($this->timeout / 10)->request(Client::METHOD_GET, $this->feed_url, $headers);
```
Art4, line 1749: `$this->timeout/10` could also lead to `float` instead of `int`.

jtojnar: Thanks, fixed.

tests/Integration/SimplePieTest.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+// SPDX-FileCopyrightText: 2004-2023 Ryan Parman, Sam Sneddon, Ryan McCue
+// SPDX-FileCopyrightText: 2023 Art4 <art4@wlabs.de>
```
jtojnar, line 4: @Art4  I can remove your credit if you want.

Although, according to [REUSE Specification – Version 3.0](https://reuse.software/spec/) _Copyright and Licensing Information_ is “the information that lists the copyright holders of a file […]” and, as I understand it, since SimplePie has no copyright assignment, you retain the copyright according to [Berne Convention](https://en.wikipedia.org/wiki/Berne_Convention).

Art4: Yes, please remove my username and the email from the source code. I do not care much about to the mention of my name.

src/SimplePie.php
```diff
@@ -3202,6 +3216,23 @@ private function get_cache(string $feed_url = ''): DataCache
 
         return $this->cache;
     }
+
+    /**
+     * Get a HTTP client
+     */
+    private function get_http_client(): Client
+    {
+        return new FileClient(
```
jtojnar, line 3225: Now that we have same timeout, we can cache the client without worries.

src/Sanitize.php
```diff
@@ -612,6 +633,26 @@ private function get_cache(string $image_url = ''): DataCache
...
         return $this->cache;
     }
+
+    /**
+     * Get a HTTP client
+     */
+    private function get_http_client(): Client
+    {
+        if ($this->http_client === null) {
+            return new FileClient(
```
jtojnar, line 647: Similarly here.

Would be nice to merge https://github.com/simplepie/simplepie/pull/817 so that the methods can be the same.

jtojnar:
Rebased.

src/File.php
```diff
@@ -390,27 +368,50 @@ public function get_body_content(): string
         return (string) $this->body;
     }
 
-    private function get_from_headers(string $name): array
+    /**
+     * Check if the $headers property was changed and update the internal state accordingly.
+     */
+    private function maybe_update_headers(): void
```
Art4, line 374: I'm not sure why we need this? Why not just use the previous `get_from_headers()` method?

src/HTTP/Parser.php
```diff
@@ -232,9 +240,17 @@ protected function new_line()
             $this->name = strtolower($this->name);
             // We should only use the last Content-Type header. c.f. issue #1
             if (isset($this->headers[$this->name]) && $this->name !== 'content-type') {
-                $this->headers[$this->name] .= ', ' . $this->value;
+                if ($this->psr7Compatible) {
```
Art4, line 243: We should create tests for this.

src/Locator.php
```diff
@@ -199,13 +200,13 @@ protected function search_elements_by_tag($name, &$done, $feeds)
...
                     try {
-                        $response = $this->get_http_client()->request(Client::METHOD_GET, $href, $headers);
-                    } catch (HttpException $th) {
-                        continue;
-                    }
+                        $feed = $this->get_http_client()->request(Client::METHOD_GET, $href, $headers);
 
-                    if ((! preg_match('/^http(s)?:\/\//i', $response->get_requested_uri()) || ($response->get_status_code() === 200 || $response->get_status_code() > 206 && $response->get_status_code() < 300)) && $this->is_feed($response, true)) {
-                        $feeds[$href] = $response;
+                        if ((!Misc::is_remote_uri($feed->get_final_requested_uri()) || ($feed->get_status_code() === 200 || $feed->get_status_code() > 206 && $feed->get_status_code() < 300)) && $this->is_feed($feed, true)) {
```
Art4, line 205: We should move this line outside of the `try` block because we don't expect it to throw a `HttpException`.

src/Sanitize.php
```diff
@@ -652,20 +647,30 @@ private function get_cache(string $image_url = ''): DataCache
...
-            return new FileClient(
+            $this->http_client = new FileClient(
                 $this->registry,
                 [
                     'timeout' => $this->timeout,
                     'redirects' => 5,
                     'useragent' => $this->useragent,
                     'force_fsockopen' => $this->force_fsockopen,
-                    'curl_options' => $this->curl_options,
+                    'curl_options' =>  $this->curl_options,
```
Art4, line 657:
```suggestion
                    'curl_options' => $this->curl_options,
```

src/Sanitize.php
```diff
@@ -652,20 +647,30 @@ private function get_cache(string $image_url = ''): DataCache
     private function get_http_client(): Client
     {
         if ($this->http_client === null) {
-            return new FileClient(
+            $this->http_client = new FileClient(
```
Art4, line 650: It's possible to change the parameters with `pass_file_data()` multiple times. Caching the FileClient would not reflect this changes in the FileClient and is a BC break. Alternatively we could unset $this->http_client inside `pass_file_data()`.

src/Locator.php
jtojnar: I moved it inside to be skipped when the exception is raised to fix the issue mentioned in https://github.com/simplepie/simplepie/pull/774#discussion_r1141241016.

I guess we could keep it outside of try and duplicate the `$done[] = $href;` assignment in the `catch` clause instead but neither solution feels particularly elegant to me.

src/Sanitize.php
jtojnar: Good catch, did the latter.

jtojnar (commented):
Thanks for the review, fixed.

src/File.php
Art4: Thank you for the explanation and for your attention to this detail.

src/HTTP/Parser.php
Art4: Yes, thank you 👍

src/Sanitize.php
Art4: We have the same problem with class `SimplePie` and the methods `set_timeout()`, `set_curl_options()`, `force_fsockopen()` and `set_useragent()`. I have solved this issue in #777 by unset the http_client or trigger an error, if a PSR-18 client was provided.

```php
    public function force_fsockopen(bool $enable = false)
    {
        $this->force_fsockopen = $enable;

        // Reset a possible existing FileClient,
        // so a new client with the changed value will be created
        if (is_object($this->http_client) && $this->http_client instanceof FileClient) {
            $this->http_client = null;
        } elseif (is_object($this->http_client)) {
            // Trigger notice if a PSR-18 client was set
            trigger_error(sprintf(
                'Using "%s()" has no effect, because you already provided a HTTP client with "%s::set_http_client()". Configure fsockopen in your HTTP client instead.',
                __METHOD__,
                get_class($this)
            ), \E_USER_NOTICE);
        }
    }
```

jtojnar: Thanks, I made it a hard error since here user would have to explicitly switch to `set_http_client`.

src/Misc.php
```diff
@@ -50,6 +50,11 @@ public static function absolutize_url($relative, $base)
         return $iri->get_uri();
     }
 
+    public static function is_remote_uri(string $uri): bool
```
Art4, line 53: We should add a note at the `CHANGELOG.md` about this new method.

src/Sanitize.php
Art4: Fair enough. 😊 
But we has also to unset a possible created $this->http_client on calling` set_timeout()`, `set_curl_options()`, etc.

src/SimplePie.php
```diff
@@ -804,6 +812,14 @@ public function set_raw_data(string $data)
...
     public function set_timeout(int $timeout = 10)
     {
+        if ($this->http_client_injected) {
+            throw new SimplePieException(sprintf(
+                'Using "%s()" has no effect, because you already provided a HTTP client with "%s::set_http_client()". Configure timeout in your HTTP client instead.',
+                __METHOD__,
+                self::class
+            ));
+        }
+
```
Art4, line 822: We should unset a possible created $this->http_client on calling` set_timeout()`, `set_curl_options()`, etc.
```suggestion

        // Reset a possible existing FileClient,
        // so a new client with the changed value will be created
        $this->http_client = null;
```

src/SimplePie.php
```diff
@@ -817,6 +833,14 @@ public function set_timeout(int $timeout = 10)
...
     public function set_curl_options(array $curl_options = [])
     {
+        if ($this->http_client_injected) {
+            throw new SimplePieException(sprintf(
+                'Using "%s()" has no effect, because you already provided a HTTP client with "%s::set_http_client()". Configure custom curl options in your HTTP client instead.',
+                __METHOD__,
+                self::class
+            ));
+        }
+
```
Art4, line 843: We should unset a possible created $this->http_client on calling `set_curl_options()`.
```suggestion

        // Reset a possible existing FileClient,
        // so a new client with the changed value will be created
        $this->http_client = null;
```

src/SimplePie.php
```diff
@@ -828,6 +852,14 @@ public function set_curl_options(array $curl_options = [])
...
     public function force_fsockopen(bool $enable = false)
     {
+        if ($this->http_client_injected) {
+            throw new SimplePieException(sprintf(
+                'Using "%s()" has no effect, because you already provided a HTTP client with "%s::set_http_client()". Configure fsockopen in your HTTP client instead.',
+                __METHOD__,
+                self::class
+            ));
+        }
+
```
Art4, line 862: We should unset a possible created $this->http_client on calling `force_fsockopen()`.
```suggestion

        // Reset a possible existing FileClient,
        // so a new client with the changed value will be created
        $this->http_client = null;
```

src/SimplePie.php
```diff
@@ -1259,6 +1291,14 @@ public function set_source_class(string $class = Source::class)
...
     public function set_useragent(?string $ua = null)
     {
+        if ($this->http_client_injected) {
+            throw new SimplePieException(sprintf(
+                'Using "%s()" has no effect, because you already provided a HTTP client with "%s::set_http_client()". Configure user agent string in your HTTP client instead.',
+                __METHOD__,
+                self::class
+            ));
+        }
+
```
Art4, line 1301: We should unset a possible created $this->http_client on calling `set_useragent()`.
```suggestion

        // Reset a possible existing FileClient,
        // so a new client with the changed value will be created
        $this->http_client = null;
```

src/SimplePie.php
jtojnar: The code will throw an exception so the value should not be changed, though.

Art4: Oh, I see... 🤦

src/Misc.php
jtojnar: Good point. I have decided to mark it as internal instead to avoid increasing the API surface and having to deprecate it later.

Art4 (approved):

Art4:
@mblaney This is ready for merge.

PR #819, opened 2023-03-19
tests/Integration: Add FileTest

jtojnar:
This is a regression test for https://github.com/simplepie/simplepie/pull/812.


jtojnar:
@Art4 Do you think something like this is reasonable?

I tried finding a library for the PHP server but after reviewing the following, I gave up and wrote my own helper class:

* https://packagist.org/packages/internations/http-mock
  - no PHP 8 support
  - incompatible dependencies
  - no open port selection
  - unmaintained
* https://packagist.org/packages/pagely/http-mock
  - fork of the above
  - no open port selection
  - still incompatible dependencies
  - unmaintained
* https://packagist.org/packages/elnoro/php-http-mock
  - 👍 minimal codebase
  - requires PHP 8
  - no open port selection
  - no released versions
* https://github.com/LibreSign/behat-builtin-extension
  - 👍 minimal codebase
  - 👍 open port selection
  - undeclared behat dependency
  - AGPL licensed
* https://packagist.org/packages/mcustiel/phiremock-server
  - complex configuration
* https://packagist.org/packages/ciareis/bypass
  - old version still supports PHP 7.2
    - but port autodetection only works on PHP 8 (using `php -S localhost:` flag)

Art4:
I would have said that it is not worth the effort to create such a test because the File class will be deprecated. But since you've already done the work, we can keep the test. Requiring a separate library for it would definitely be too much.

tests/Integration/Fixtures/redirects.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+function redirect(int $status, string $location): void
```
Art4, line 5: Can we use anonymous function instead of populating the global namespace?
```suggestion
$redirect = function(int $status, string $location): void
```

tests/Integration/Fixtures/redirects.php
```diff
@@ -0,0 +1,39 @@
...
+{
+    http_response_code($status);
+    header("Location: $location");
+}
```
Art4, line 9:
```suggestion
};
```

Art4 (changes requested):

tests/Integration/Fixtures/redirects.php
jtojnar: Good point. I have moved it to a separate file as per PSR-1:

> Files SHOULD either declare symbols (classes, functions, constants, etc.) or cause side-effects (e.g. generate output, change .ini settings, etc.) but SHOULD NOT do both.

jtojnar (commented):
Thanks for the review, updated. I think this integration test will be useful for PSR-18 as well since how the permanent/temporary redirection interact with `get_final_url` is non-trivial. Though there we at least will not need to spawn a server since that will support mocking.

Art4:
I've found this library for mocking a webserver: https://packagist.org/packages/donatj/mock-webserver

- 👍 still supports PHP 7.1 up to PHP 8.4
- 👍 open for port and host selection
- 👍 maintained
- all (only 1) dependencies are compatible
- MIT licence

I've looked into the code and it looks very good and straight forward. I've added it in #905.

jtojnar:
This was superseded by  https://github.com/simplepie/simplepie/pull/909.

jtojnar:
The test cases are ported to the new infrastructure in #918.

PR #822, opened 2023-03-19
tests: Skip cases incompatible with PHP 8

jtojnar:
instead of running them and expecting an error.

Arguably, it is semantically cleaner to avoid testing a behaviour that is defined by PHP itself rather than anything SimplePie is doing.

As a bonus, it gets rid of the following warning:

    Expecting E_ERROR and E_USER_ERROR is deprecated and will no longer be possible in PHPUnit 10.

### DIFF
--- a/tests/CacheTest.php
+++ b/tests/CacheTest.php
@@ -41,14 +41,14 @@ class CacheTest extends PHPUnit\Framework\TestCase
 
     public function testDirectOverrideLegacy()
     {
-        if (version_compare(PHP_VERSION, '8.0', '<')) {
-            $this->expectException(SuccessException::class);
-        } else {
+        if (version_compare(PHP_VERSION, '8.0', '>=')) {
             // PHP 8.0 will throw a `TypeError` for trying to call a non-static method statically.
             // This is no longer supported in PHP, so there is just no way to continue to provide BC
             // for the old non-static cache methods.
-            $this->expectError();
+            $this->markTestSkipped('Using legacy cache class is not compatible with PHP 8.');
         }
+
+        $this->expectException(SuccessException::class);
 
         $feed = new SimplePie();
         $this->expectDeprecation();
--- a/tests/Unit/SimplePieTest.php
+++ b/tests/Unit/SimplePieTest.php
@@ -225,20 +225,20 @@ class SimplePieTest extends TestCase
 
     public function testLegacyCallOfSetCacheClass()
     {
+        if (version_compare(PHP_VERSION, '8.0', '>=')) {
+            // PHP 8.0 will throw a `TypeError` for trying to call a non-static method statically.
+            // This is no longer supported in PHP, so there is just no way to continue to provide BC
+            // for the old non-static cache methods.
+            $this->markTestSkipped('Using legacy cache class is not compatible with PHP 8.');
+        }
+
         $feed = new SimplePie();
         $this->expectDeprecation();
         $feed->set_cache_class(LegacyCacheMock::class);
         $feed->get_registry()->register(File::class, FileMock::class);
         $feed->set_feed_url('http://example.com/feed/');
 
-        if (version_compare(PHP_VERSION, '8.0', '<')) {
-            $this->expectException(SuccessException::class);
-        } else {
-            // PHP 8.0 will throw a `TypeError` for trying to call a non-static method statically.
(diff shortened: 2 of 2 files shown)


jtojnar:
cc @jrfnl who introduced this in https://github.com/simplepie/simplepie/pull/691.

jrfnl (commented):
@jtojnar I agree these tests will need to be skipped when the tests start running on PHPUnit 10.

However...
1. SimplePie is not using PHPUnit 10 (yet).
    The current warning doesn't do any harm.
2. The test would now be skipped on PHP >= 8.0, while on PHP 8.0 (which would be using PHPUnit 9.x even when PHPUnit 10 would start to be allowed), it can still run.
    I.e. it is now being skipped on more PHP versions then needed.
3. The current change doesn't include a "skip reason", which is recommended to allow people running the tests in `verbose` mode to quickly see why tests are being skipped.

I'd recommend using the [`@requires PHPUnit < 10` docblock annotation](https://docs.phpunit.de/en/9.6/annotations.html#requires) instead.

Aternatively the PHPUnit 10 attribute could be used: [`#[RequiresPhpunit('< 10.0.0')]`.](https://docs.phpunit.de/en/main/attributes.html#requiresphpunit) though that may run into "Unknown class" errors ?

Whether the annotation or the attribute is used, in both cases a "skip reason" will be shown and the test should still run on PHP 8.0.

---

As a side note: there is no PHPUnit 10.x compatible release (yet) for the [PHPUnit Polyfills](https://github.com/Yoast/PHPUnit-Polyfills) (used under the hood to allow the tests to run on multiple PHP/PHPunit versions).
A new release which supports PHPUnit 10.x is being worked on and the roadmap has been published and is open for comments: https://github.com/Yoast/PHPUnit-Polyfills/issues/95

Art4 (approved):

jtojnar:
>     1. SimplePie is not using PHPUnit 10 (yet).
>        The current warning doesn't do any harm.

It is a bit annoying, though the primary reason I went for this change is that it felt weird essentially testing a behaviour that is defined by PHP itself rather than anything SimplePie is doing.

>     2. The test would now be skipped on PHP >= 8.0, while on PHP 8.0 (which would be using PHPUnit 9.x even when PHPUnit 10 would start to be allowed), it can still run.
>        I.e. it is now being skipped on more PHP versions then needed.

I am not sure why it could run on PHP 8.0, it looks like the original version check was correct according to https://php.watch/versions/8.0/non-static-static-call-fatal-error and I just inverted it.

>     3. The current change doesn't include a "skip reason", which is recommended to allow people running the tests in `verbose` mode to quickly see why tests are being skipped.

Thanks. Added.

> I'd recommend using the [`@requires PHPUnit < 10` docblock annotation](https://docs.phpunit.de/en/9.6/annotations.html#requires) instead.

I would probably wait for PHPUnit 10.x compatible release of [PHPUnit Polyfills](https://github.com/Yoast/PHPUnit-Polyfills).

jtojnar:
Oh, that makes sense. I assumed it was more like [Symfony’s PHPUnit Bridge](https://github.com/symfony/phpunit-bridge).

If the functionality will be removed, would not `#[RequiresPhpunit('< 10.0.0')]` be redundant? The version of the Polyfills that will support PHPUnit 10 would not have the `Yoast\PHPUnitPolyfills\Polyfills\ExpectPHPException` trait in the first place.

jrfnl:
> Oh, that makes sense. I assumed it was more like [Symfony’s PHPUnit Bridge](https://github.com/symfony/phpunit-bridge). I like the Polyfills better then.
> 
> If the functionality will be removed, would not `#[RequiresPhpunit('< 10.0.0')]` be redundant? The version of the Polyfills that will support PHPUnit 10 would not have the `Yoast\PHPUnitPolyfills\Polyfills\ExpectPHPException` trait in the first place.

Depends on whether or not a Helper class for transitioning will be added to the Polyfills. The trait will definitely be removed, but I haven't seen anyone give an opinion yet on whether or not they think having a Helper class (to selectively use) would be helpful.

My own biggest concern with this change in PHPUnit is that packages can no longer deprecate functionality + test the deprecation notice. I think that will be (is) the biggest point of friction.

Art4:
We are now using PHPUnit 9+10 and have removed the dependency to Yoast\PHPUnitPolyfills, see #848.

PR #827, opened 2023-03-20
Item: Accept Sanitize object

jtojnar:
`SimplePie::$sanitize` is a private member so we should not access it.
It only works because the visibility is currently described using PHPDoc annotation.


jtojnar:
@mblaney it would probably be better to rebase merge rather than squash merge this, since doing the latter makes it harder to reason about the code when looking at the history. :woman_shrugging:

PR #834, opened 2023-05-01
Typehinting

ColonelMoutarde:
#### When submitting a pull request

1. Run the linter.

   ```bash
   make lint
   ```

1. Run the tests.

   ```bash
   make test
   ```

1. Carefully _read your diff_ before submitting a PR.
    * Make sure your diff clearly represents your changes.
    * Make sure your code is easy for reviewers to follow.
    * Can your code review be broken into smaller chunks?
    * Make sure you have PHPUnit tests that cover your changes.

1. Make sure that your changes are very-nearly complete.
    * Small changes/fixes based on feedback are OK.
    * Major changes should be avoided while a PR is open. If you need to make major changes, close the PR, make your changes, and open a new PR once you're ready.

1. In the description, start with an explanation of the change and why you want to make it.
    * Make any relevant documentation easily available for reviewers.

1. Add a list of changes you've made using [task list](https://help.github.com/articles/basic-writing-and-formatting-syntax/#task-lists) syntax.
    * Things that are complete should have `[X]`.
    * Things that are still incomplete should have `[ ]`.

1. Discuss feedback.
    * The _SimplePie NG Core Team_ has the final say on what gets accepted.
    * Respond to all code review feedback.
    * _Respectfully_ discuss any feedback you disagree with. You may need to bring a _better_ argument to the table to convince us.

1. If you don't understand something, _ask_.

### DIFF
--- a/src/Misc.php
+++ b/src/Misc.php
@@ -41,7 +41,7 @@ class Misc
         return $time;
     }
 
-    public static function absolutize_url($relative, $base)
+    public static function absolutize_url($relative, $base): string
     {
         $iri = \SimplePie\IRI::absolutize(new \SimplePie\IRI($base), $relative);
         if ($iri === false) {
@@ -56,9 +56,8 @@ class Misc
      * @deprecated since SimplePie 1.3, use DOMDocument instead (parsing HTML with regex is bad!)
      * @param string $realname Element name (including namespace prefix if applicable)
      * @param string $string HTML document
-     * @return array
      */
-    public static function get_element(string $realname, string $string)
+    public static function get_element(string $realname, string $string): array
     {
         // trigger_error(sprintf('Using method "' . __METHOD__ . '" is deprecated since SimplePie 1.3, use "DOMDocument" instead.'), \E_USER_DEPRECATED);
 
@@ -89,7 +88,7 @@ class Misc
         return $return;
     }
 
-    public static function element_implode($element)
+    public static function element_implode(array $element): string
     {
         $full = "<$element[tag]";
         foreach ($element['attribs'] as $key => $value) {
@@ -140,7 +139,7 @@ class Misc
         return $message;
     }
 
-    public static function fix_protocol($url, $http = 1)
+    public static function fix_protocol(string $url, $http = 1); string
     {
         $url = Misc::normalize_url($url);
         $parsed = Misc::parse_url($url);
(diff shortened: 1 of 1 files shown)


jtojnar:
This will be a BC break.

src/Misc.php
```diff
@@ -140,7 +139,7 @@ public static function error($message, $level, $file, $line)
         return $message;
     }
 
-    public static function fix_protocol($url, $http = 1)
+    public static function fix_protocol(string $url, $http = 1); string
```
jtojnar, line 142: Here you have semicolon.

Maikuolan:
> This will be a BC break.

^ That.

Specifically, support for return type declarations wasn't added until PHP 7.0, so adding them to SimplePie would, at the very least, necessitate increasing the minimum required PHP version as stated at the repository's README and `composer.json` files from 5.3 to 7.

*See*:
- *https://www.php.net/manual/en/migration70.new-features.php*
- *https://wiki.php.net/rfc/return_types*

jtojnar:
> necessitate increasing the minimum required PHP version as stated at the repository's README and `composer.json` files from 5.3 to 7.

We are already targetting PHP 7.2 since SimplePie 1.8.0.

Maikuolan:
> We are already targetting PHP 7.2 since SimplePie 1.8.0.

Whoops.. So we are. Scratch my reply then. My bad. '^.^

Self-reminder: Looking at multiple different repositories/projects/etc at the same time isn't always such a great idea. SimplePie aside, there's another RSS library I was looking at which supported older PHP versions, and I think I got my wires crossed there.

Anyway, all good.

ColonelMoutarde:
All fixed :°)

ColonelMoutarde:
All tests success

ColonelMoutarde:
> I'm sorry, but as @jtojnar already [said](https://github.com/simplepie/simplepie/pull/834#issuecomment-1529932435): This will be a BC break. Users who had extend e.g. `Misc::fix_protocol()` will get an exception.

Do you have an suggestion ?

Art4:
> Do you have an suggestion ?

Only suggestion is to defer your PR until we are working at release v2. Increasing the major version allows breaking changes.

But we are already planning to add return types by introducing new camelCase methods starting with 2.1.0, see #731.

PR #835, opened 2023-05-19
Move code style check into own Github Actions job

Art4:
This PR will move the code style check in CI (introduced with #825) to a separate job. The reason for this is that it is confusing when the tests fail only for a specific PHP version, see https://github.com/simplepie/simplepie/actions/runs/5021614275 as an example.

With this PR the code style check is done in a separate job, so an error "only" in the code style can be visually detected as not serious.

This PR also runs the tests for the single-file build under PHP 8.2 and indicates this in the job name.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,13 @@ jobs:
   test-compiled:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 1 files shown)


jtojnar (approved):
LGTM.

PR #756, opened 2022-10-23
Add docs into the repo

jtojnar:
I downloaded the HTML from live site and created a script (`markdownify.py`) that will convert them to Markdown. This PR attempts to make everything work. We will remove/update the outdated content in a follow-up PR.

To regenerate the markdown files, run `python3 markdownify.py` in the SimplePie branch after installing `python3-beautifulsoup4`, `python3-pypandoc` and `prettier`. Or if you have [Nix](https://nixos.org/), you can just run `nix-shell -I 'nixpkgs=channel:nixos-unstable' -p 'python3.withPackages (ps: with ps; [ beautifulsoup4 pypandoc ])' nodePackages.prettier --run 'python3 markdownify.py'`.

Currently I am using [Zola](https://www.getzola.org/) as the generator, as I am most familiar with it. A different SSG can be used if preferred.

To preview run `zola serve` in the `docs` subdirectory.

## TODO

- [x] Have index auto-generated for blog.
- [x] Move `h1` out of individual pages into template.
- [x] Strip blog junk.
- [x] Strip wiki junk.
- [x] Ensure links are correct.
- [x] Figure out what to do with the demo.
- [x] Add README to the `docs/` directory.
- [x] Remove the old files.
- [x] Set up ApiGen.
- [x] Set up GitHub pages CI.
    - https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/managing-a-custom-domain-for-your-github-pages-site

Fixes: #543

PR #836, opened 2023-05-22
docs: Add docs into the repo

jtojnar:
Second attempt at using static site generator for docs, this time without wiki.

Supersedes: https://github.com/simplepie/simplepie/pull/756


jtojnar:
I am still not sure if having the website in-tree is the right move. Currently, I am leaning towards the website + blog in separate repo.

jtojnar:
Let’s. We can always move it back if we have some time left.

PR #838, opened 2023-05-31
SimplePie: Prepare PubSubHubbub for PSR-18 clients

jtojnar:
Previously, this only supported `File`, let’s prepare for other `Response` implementations.

Also fixes a bug where it would clear previous links (by setting it to comma instead of appending).
And where it would add redundant `rel=self` link.

### DIFF
--- a/src/HTTP/Response.php
+++ b/src/HTTP/Response.php
@@ -95,7 +95,7 @@ interface Response
      *         }
      *     }
      *
-     * @return string[][] Returns an associative array of the message's headers.
+     * @return array<non-empty-array<string>> Returns an associative array of the message's headers.
      *     Each key MUST be a header name, and each value MUST be an array of
      *     strings for that header.
      */
--- a/src/HTTP/Response.php
+++ b/src/HTTP/Response.php
@@ -128,6 +128,20 @@ interface Response
     public function get_header(string $name): array;
 
     /**
+     * Return an instance with the provided value replacing the specified header.
+     *
+     * This method MUST be implemented in such a way as to retain the
+     * immutability of the message, and MUST return an instance that has the
+     * new and/or updated header and value.
+     *
+     * @param string $name Case-insensitive header field name.
+     * @param string|non-empty-array<string> $value Header value(s).
+     * @return static
+     * @throws \InvalidArgumentException for invalid header names or values.
+     */
+    public function with_header(string $name, $value);
+
+    /**
      * Retrieves a comma-separated string of the values for a single header.
      *
      * This method returns all of the header values of the given
--- a/src/SimplePie.php
+++ b/src/SimplePie.php
@@ -2032,8 +2032,9 @@ class SimplePie
                         $this->all_discovered_feeds
                     );
                     if ($microformats) {
-                        if ($hub = $locate->get_rel_link('hub')) {
-                            $self = $locate->get_rel_link('self');
+                        $hub = $locate->get_rel_link('hub');
+                        $self = $locate->get_rel_link('self');
+                        if ($hub || $self) {
                             $file = $this->store_links($file, $hub, $self);
                         }
                         // Push the current file onto all_discovered feeds so the user can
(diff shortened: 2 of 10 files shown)


src/HTTP/Response.php
```diff
@@ -127,6 +127,20 @@ public function has_header(string $name): bool;
      */
     public function get_header(string $name): array;
 
+    /**
```
jtojnar, line 130: Again, not sure if we should duplicate the comments, rather than using `{@inheritdoc}`

Art4 (approved):

Art4:
@mblaney Are there any objections?

jtojnar:
I will try to get back to SimplePie development on weekend.

mblaney:
> @mblaney Are there any objections?

no this looks good to me :+1:

jtojnar:
Turns out `SimplePie::get_links()` does not consider `link` elements from HTML files so the `store_links()` method is still needed even if we do not want to support `a[rel]`.

Added tests, split the fixes into separate commits, and changed `get_headers` to return array of non-empty-array of string.

jtojnar:
Also removing support for `a[rel]`.

tests/Integration/SimplePieTest.php
```diff
@@ -253,4 +253,252 @@ public function testRequestingAFeedFromPsr16CacheWorks(): void
...
+        <link rel="hub" href="https://pubsubhubbub.appspot.com">
+        <link rel="self" href="https://example.com">
+        <div class="h-feed">
+        </div>
+    </body>
+</html>
+HTML
+            ,
+            'hubUrl' => 'https://pubsubhubbub.appspot.com/',
+            'selfUrl' => 'https://example.com/',
```
Art4, line 404: `rel=hub` only from `<body>` section should not be supported.

```suggestion
            'hubUrl' => null,
            'selfUrl' => null',
```

tests/Unit/ParserTest.php
```diff
@@ -29,36 +29,7 @@ public function testClassExists(): void
...
-                <li class="h-entry">
-                    <a href="/4" class="u-url p-name">Four</a>
-                </li>
-            </ul>
-        </div>
-    </body>
-</html>
-HTML
-            ,
+            file_get_contents(dirname(__FILE__, 2) . '/data/microformats/h-feed-simple.html'),
```
Art4, line 32: Can we keep the heredoc for better readability and just indent it properly?

```suggestion
            <<<HTML
            <!-- SPDX-FileCopyrightText: 2018 Aaron Parecki -->
            <!-- SPDX-License-Identifier: CC0-1.0 -->
            <!-- SPDX-ArtifactOfProjectName: php-mf2 -->
            <html>
                <head>
                    <title>Test</title>
                </head>
                <body>
                    <div class="h-feed">
                        <a href="/author" class="p-author h-card">Author Name</a>
                        <ul>
                            <li class="h-entry">
                                <a href="/1" class="u-url p-name">One</a>
                            </li>
                            <li class="h-entry">
                                <a href="/2" class="u-url p-name">Two</a>
                            </li>
                            <li class="h-entry">
                                <a href="/3" class="u-url p-name">Three</a>
                            </li>
                            <li class="h-entry">
                                <a href="/4" class="u-url p-name">Four</a>
                            </li>
                        </ul>
                    </div>
                </body>
            </html>
            HTML
            ,
```

Art4 (changes requested):

src/SimplePie.php
```diff
@@ -2034,9 +2034,7 @@ protected function fetch_data(&$cache)
                     if ($microformats) {
                         if ($hub = $locate->get_rel_link('hub')) {
                             $self = $locate->get_rel_link('self');
```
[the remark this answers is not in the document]
jtojnar: As described in <https://www.w3.org/TR/websub/#x8-1-discovery>:

> The decision about whether a subscriber should look for `<link>` elements inside a page's `<body>` (as well as the `<head>`) is not straightforward, and there is currently no clear consensus.

tests/Integration/SimplePieTest.php
Art4: Difficult topic.. Can we at least make sure that the following example results into `example.com` instead of `evil.example.com`?

```<html>
    <head>
        <title>Test</title>
        <link rel="hub" href="https://pubsubhubbub.appspot.com">
        <link rel="self" href="https://example.com">
    </head>
    <body>
        <link rel="hub" href="https://pubsubhubbub.appspot.com">
        <link rel="self" href="https://evil.example.com">
        <div class="h-feed">
        </div>
    </body>
</html>```

jtojnar:
This now also appears to be blocked by #921

Art4 (approved):
Thanks for the great work. 👍

Issue #830, opened 2023-04-18
[BUG] Cache constantly updating


Sir-Will:
#### Description

It doesn't seem like the cache is working correctly for me. Every time I reload the page, I see the timestamps of the cache files update.
This shouldn't be the case, or?

#### Steps to Reproduce

```php
        $infoFeed = new SimplePie();
        $infoFeed->set_feed_url($infoFeeds);
        $infoFeed->set_cache_duration(600);
        $infoFeed->init();
        $infoFeed->handle_content_type();
```

#### Expected Behavior

Cache not updating every run.

#### Actual Behavior

Cache is updating on every run.

#### Possible Solutions

If you have already ideas how to solve the issue, add them here. (remove this section if not needed)

#### Environment

PHP 8.1
SimplePie 1.8.0

bart-v:
Same problem here.

This is caused by these 4 occurrences of `cache_expiration_time`
https://github.com/simplepie/simplepie/blob/master/src/SimplePie.php#L1676
https://github.com/simplepie/simplepie/blob/master/src/SimplePie.php#L1769
https://github.com/simplepie/simplepie/blob/master/src/SimplePie.php#L1920

`cache_expiration_time` => should be => `__cache_expiration_time`

Or at least the array index should be consistent everywhere...

Art4:
The default file cache is deprecated an will be removed in the next major release. Instead you should use a PSR-16](https://www.php-fig.org/psr/psr-16/) cache implementation like [Symfony Cache](https://symfony.com/doc/current/components/cache.html) or [every other implementation](https://packagist.org/providers/psr/simple-cache-implementation).

```php
$simplepie = new \SimplePie\SimplePie();
$simplepie->set_cache(
    new \Symfony\Component\Cache\Psr16Cache(
        new \Symfony\Component\Cache\Adapter\FilesystemAdapter()
    ),
);
```

bart-v:
Then, a simple example on how to then set the cache type (i.e. file), location and duration would be nice.
The current documentation only mentions a deprecated example and does not say anything about PSR-16 ...

bart-v:
At least it's clear that the PSR-16 changes in https://github.com/simplepie/simplepie/pull/752 broke the legacy cache.
I think reverting https://github.com/simplepie/simplepie/pull/752/commits/3ba7c912d470f5bf9ddf2e414430e52d2f6a9823 will fix this bug
Would you consider this, please?

Art4:
[...]

PR #845, opened 2023-08-18
Make array index "__cache_expiration_time" consistent everywhere

bart-v:
This fixes https://github.com/simplepie/simplepie/issues/830


Art4:
`__cache_expiration_time` is only used inside deprecated `src/Cache/BaseDataCache.php` for BC and should not be used outside of the class.

PR #846, opened 2023-08-28
Stop legacy file cache constantly updating

### LINKED ISSUE #830: [BUG] Cache constantly updating

Art4:
This PR fixes a bug introduced in #752.

Fixes #830.


Art4:
Something very strange is going on here. Builds [606](https://github.com/simplepie/simplepie/actions/runs/5997049083) and [608](https://github.com/simplepie/simplepie/actions/runs/5997307058) are for the exact same code changes but the results for different PHP versions is different.

grafik

I can not reproduce this on my machine (tests are successfully on PHP 8.2). I assume some race conditions on the Github Actions?

Art4:
[I can reproduce this strange behavior on my GitHub fork](https://github.com/Art4/simplepie/actions/runs/5996993450/attempts/4): I have rerun the job for the same commit 4 times and got randomly different errors each time.

grafik

jtojnar:
I can reproduce it locally. Will try to debug it during the week.

src/SimplePie.php
```diff
@@ -1901,7 +1901,7 @@ protected function fetch_data(&$cache)
                     $this->data = [];
                 }
                 // Check if the cache has been updated
-                elseif (isset($this->data['cache_expiration_time']) && $this->data['cache_expiration_time'] > time()) {
+                elseif (isset($this->data['cache_expiration_time']) && $this->data['cache_expiration_time'] < time()) {
```
Alkarex, line 1904:
```suggestion
                elseif (empty($this->data['cache_expiration_time']) || $this->data['cache_expiration_time'] < time()) {
```

Otherwise you hit the next `else` which believes the cache to be still valid

Alkarex: Note that the equality here gives problem when tests are running within the same second, which is the reason for your tests failing.

https://github.com/simplepie/simplepie/blob/9385c54382edec0a768fdbe310c3e580c1b5dc81/tests/Integration/CachingTest.php#L123-L124

You can for instance change to `$defaultMtime = time() - 1`;

See https://github.com/FreshRSS/simplepie/pull/23

Alkarex: Replacement PR with additional fixes https://github.com/simplepie/simplepie/pull/883
Feel free to merge in this PR or in master as you prefer

tests/Integration/CachingTest.php
```diff
@@ -246,10 +246,10 @@ public static function provideSavedCacheData(): array
             [CacheInterface::class, $currentlyCachedDataWithNonFeedUrl,     $expectDataWithNewFeedUrl, $defaultMtime],
             // Check if the cache has been updated
             [Base::class,           $currentlyCachedDataIsUpdated,          $expectDefaultDataWritten, $defaultMtime],
-            [CacheInterface::class, $currentlyCachedDataIsUpdated,          $expectDefaultDataWritten, $defaultMtime],
+            [CacheInterface::class, $currentlyCachedDataIsUpdated,          $expectNoDataWritten,      $defaultMtime],
```
jtojnar, line 249: Any idea why this differs from the legacy backend?

src/SimplePie.php
jtojnar: > Otherwise you hit the next `else` which believes the cache to be still valid

I would expect that when no expiration time is provided, it cannot really expire so this is correct IMO. Though it [should not actually happen](https://github.com/simplepie/simplepie/pull/883#discussion_r1776094433) if I read the code correctly so the `isset` is there mostly for PHPStan.

Alkarex: If `cache_expiration_time` is not set for any reason, I would want my cache to be revalidated. Otherwise it can stay like that for a long time

jtojnar: Fair enough. I still wonder why it duplicates the ttl handling, though.

src/SimplePie.php
jtojnar, line 1: @Art4 Why in  bb38d740c12a953dce4e2722a8d26d7a3ed2be3f did you start to set `cache_expiration_time` as part of the data in the first place? Is it to handle implementations that do not support TTL?

[PSR 16](https://www.php-fig.org/psr/psr-16/) is inconsistent whether that is required:

> Implementing Libraries MUST support at minimum TTL functionality as described below with whole-second granularity.
> \[…\]
> Implementing Libraries MAY expire an item before its requested Expiration Time, but MUST treat an item as expired once its Expiration Time is reached.

vs.

> If the underlying implementation does not support TTL, the user-specified TTL MUST be silently ignored.

Also, unlike `getMultiple`, the docstring for `get` does not say whether stale/expired items should be returned.

Alkarex: A little note from my side: HTTP compliance and compatibility with existing SimplePie code bases should be much more important than whatever PSR 16 might be saying

jtojnar: Well, if PSR 16 implementations all support TTL, we can remove the handling from everywhere except the legacy shim.

Art4: > @Art4 Why in [bb38d74](https://github.com/simplepie/simplepie/commit/bb38d740c12a953dce4e2722a8d26d7a3ed2be3f) did you start to set `cache_expiration_time` as part of the data in the first place? Is it to handle implementations that do not support TTL?
>
> [PSR 16](https://www.php-fig.org/psr/psr-16/) is inconsistent whether that is required:
> 
> > Implementing Libraries MUST support at minimum TTL functionality as described below with whole-second granularity.
> > […]
> > Implementing Libraries MAY expire an item before its requested Expiration Time, but MUST treat an item as expired once its Expiration Time is reached.
> 
> vs.
> 
> > If the underlying implementation does not support TTL, the user-specified TTL MUST be silently ignored.

You are mixing up "Implementing Libraries" (`Psr\SimpleCache\CacheInterface` implementations) and "underlying implementation" (`SimplePie\Cache\BaseDataCache`).

`cache_expiration_time` was added to mimic the functionality or `mtime()` and `touch()` on the SimplePie side. In the `SimplePie\Cache\BaseDataCache` the TTL was implemented using an internal `__cache_expiration_time` key.

Alkarex: We have already had a similar discussion https://github.com/FreshRSS/simplepie/pull/11/files#r1725726684
My impression is that there is a combination of bugs and more than one misunderstanding (very possibly one of them being from my side :-))
I would though like to see the straightforward https://github.com/simplepie/simplepie/pull/883 (or this one, but less ready) PR merged before embarking on tackling those issues

Art4: > But the thing that was confusing me most is that we want the cache to return expired entries, which Psr16 cache does not support. I think I finally understand that after digging into [#883 (comment)](https://github.com/simplepie/simplepie/pull/883#discussion_r1777837083)

Yes, that might be the biggest issue. Maybe atm the best would be to pass `null` (unlimited) as TTL, as you suggested [here](https://github.com/simplepie/simplepie/pull/883#discussion_r1777837083) and only work with `$data['cache_expiration_time']`.

Art4:
Replaced by #883.

PR #849, opened 2023-09-13
Replace Locator and Content\Type\Sniffer with Content\Detector

Art4:
In #731 we plan to deprecate the `SimplePie\File` class. The file class is required for `SimplePie\Locator` and `SimplePie\Content\Type\Sniffer`, so the plan is to deprecate this classes too. See also the discussion in https://github.com/simplepie/simplepie/pull/826#discussion_r1210511213

This PR introduces a new interface `SimplePie\Content\Detector` that defines the most features of the Locator and Sniffer in 3 methods. 

The `SimplePie\Content\VerfiedFeedsDetector` class was created for BC. It internally uses the old Locator and Sniffer classes. This is needed because Locator and Sniffer classes could be replaced by the user using the Registry.

A new `SimplePie\Content\UnverfiedFeedsDetector` has to be created to replace the function of Locator and Sniffer. The idea is that this Detector should not have a dependency to a HTTP client but should only work with `Response` classes.

This PR is under construction and requires some other PRs like #838 and a refactoring of the microformats handling in `SimplePie\SimplePie::fetch_data()`


src/Content/VerifiedFeedsDetector.php
```diff
@@ -0,0 +1,119 @@
...
+     * @link http://web.archive.org/web/20110607232437/http://diveintomark.org/archives/2002/08/15/ultraliberal_rss_locator
+     *
+     * @param SimplePie::LOCATOR_* $discovery_level
+     *
+     * @return string[] Array of possible feed urls. The urls are not requested or checked for containing a feed
+     */
+    public function discover_possible_feed_urls(Response $response, int $discovery_level = SimplePie::LOCATOR_ALL): array
+    {
+        /** @var Locator */
+        $locator = $this->registry->create(
```
jtojnar, line 55: Unlike before, this creates a new locator entry for each method call. But it probably is not very costly and we only call the methods few times anyway.

src/SimplePie.php
```diff
@@ -2027,10 +2031,47 @@ protected function fetch_data(&$cache)
...
-                        $this->all_discovered_feeds
+                    $possible_feed_urls = $detector->discover_possible_feed_urls(
+                        $file,
+                        $this->autodiscovery
                     );
+
+                    $discovered = null;
+                    $checked_feeds = 0;
+
+                    foreach ($possible_feed_urls as $href) {
```
jtojnar, line 2042: I like having a pure (side-effect-free detector) but moving the effects to `SimplePie` class is not nice when it is already overly bloated. Could we have a separate helper class that would encapsulate this code?

src/Content/Detector.php
```diff
@@ -0,0 +1,52 @@
...
+
+declare(strict_types=1);
+
+namespace SimplePie\Content;
+
+use SimplePie\HTTP\Response;
+use SimplePie\SimplePie;
+
+/**
+ * Helper for feed auto-discovery and type sniffing
```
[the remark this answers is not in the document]
Art4: `SimplePie\Locator` internally depends on `SimplePie\Content\Type\Sniffer`, so by replacing the `Locator` class we also replace the `Sniffer` class.

PR #851, opened 2023-10-07
refactor: drop call to xml_set_object()

Girgias:
The xml_set_X_handler() functions already pass in standard callables.

Related to https://github.com/php/php-src/pull/12340


jtojnar:
Hmm, if I it is only used for `xml_set_X_handler`, it should indeed not be needed since https://github.com/simplepie/simplepie/commit/f5c99dddafe8b8478d91c94e04db0ce1f63c4401

jtojnar (approved):

Girgias:
Funny that you fixed this because of PHPStan, I reported it to them as: https://github.com/phpstan/phpstan/issues/9970

mblaney:
thanks @Girgias

PR #857, opened 2024-01-10
PHPStan level 8

grimmdude:
* Increases PHPStan to level 8.

Depends on #892, #890 , #894, #869


Art4:
Please note that we have to support PHP 7.2, so mixed and Union types are not supported there...

You should keep a look at your GitHub Action here: https://github.com/grimmdude/simplepie/actions/runs/7536994157/job/20515243462

src/Source.php
```diff
@@ -376,8 +376,8 @@ public function get_links(string $rel = 'alternate')
                     } else {
                         $this->data['links'][\SimplePie\SimplePie::IANA_LINK_RELATIONS_REGISTRY . $key] = &$this->data['links'][$key];
                     }
-                } elseif (substr($key, 0, 41) === \SimplePie\SimplePie::IANA_LINK_RELATIONS_REGISTRY) {
-                    $this->data['links'][substr($key, 41)] = &$this->data['links'][$key];
+                } elseif (substr((string) $key, 0, 41) === \SimplePie\SimplePie::IANA_LINK_RELATIONS_REGISTRY) {
```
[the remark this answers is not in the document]
grimmdude: Sure thing https://github.com/simplepie/simplepie/pull/857/commits/e52cbe6621294465d74aa9b659b2b5117eb4e06c

tests/Unit/RestrictionTest.php
```diff
@@ -90,11 +90,11 @@ public function test_get_relationship(string $data, string $expected): void
 
         if ($enclosure = $item->get_enclosure()) {
             $this->assertInstanceOf(Enclosure::class, $enclosure);
```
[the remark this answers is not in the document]
grimmdude: Here's the error that this change fixes:

```
 ------ ------------------------------------------------------------------- 
  Line   tests/Unit/RestrictionTest.php                                     
 ------ ------------------------------------------------------------------- 
  95     Cannot call method get_restriction() on SimplePie\Enclosure|null.  
 ------ ------------------------------------------------------------------- 
```

Problem was that `$restriction = $enclosure->get_restriction();` was originally outside of this conditional.

src/Parser.php
```diff
@@ -190,7 +190,7 @@ public function parse(string &$data, string $encoding, string $url = '')
                         } else {
                             $attrName = $xml->localName;
                         }
-                        $attributes[$attrName] = $xml->value;
+                        $attributes[(string) $attrName] = (string) $xml->value;
```
jtojnar, line 193: What error causes this cast? Neither `localName` or `value` are supposed to be anything but string https://www.php.net/manual/en/class.xmlreader.php

grimmdude: @jtojnar, below is the error that casting these to string resolves.  Looks like this stems from the possible types of the properties accessed from `XMLReader`.

>   Line   src/Parser.php                                                                                                                  
> 195    Parameter #3 $attributes of method SimplePie\Parser::tag_open() expects array<string, string>, array<int|string, mixed> given.

src/Parser.php
```diff
@@ -285,14 +285,19 @@ public function tag_open($parser, string $tag, array $attributes)
                 $this->xml_base_explicit[] = true;
             }
         } else {
-            $this->xml_base[] = end($this->xml_base);
+            if ($xml_base = end($this->xml_base)) {
```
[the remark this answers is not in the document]
grimmdude: Got it, just updated that thanks.

src/Parser.php
grimmdude: I've just pushed https://github.com/simplepie/simplepie/pull/857/commits/bb89c217acf9e2f3091e9820297f446f1379408f which moves this fix more upstream by only casting `localName` to string.  This also seems to fix this error 🤷‍♂️

grimmdude:
@jtojnar this is ready for review, all tests should now be passing.

tests/Unit/RestrictionTest.php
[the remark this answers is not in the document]
jtojnar: Right. But `assertInstanceOf` should fail when `null`. Maybe we need https://github.com/phpstan/phpstan-phpunit for PHPStan to properly detect that.

tests/Unit/SimplePieTest.php
```diff
@@ -206,7 +206,7 @@ public function testItemWithEmptyContent(): void
         $content = 'item description';
         $feed = $this->createFeedWithTemplate($data, $content);
         $item = $feed->get_item();
-        $this->assertSame($content, $item->get_content());
+        $this->assertSame($content, $item ? $item->get_content() : null);
```
jtojnar, line 209: Probably cleaner to do assertInstanceOf Item like we do elsewhere.

tests/Unit/LocatorTest.php
```diff
@@ -136,13 +136,16 @@ public function test_from_file(File $data): void
...
         $expected = [];
         $document = new DOMDocument();
-        $document->loadHTML($data->body);
+        $document->loadHTML((string) $data->body);
         $xpath = new DOMXPath($document);
-        foreach ($xpath->query('//link') as $element) {
-            /** @var \DOMElement $element */
-            $expected[] = 'http://example.com' . $element->getAttribute('href');
+
+        if ($queryResult = $xpath->query('//link')) {
```
jtojnar, line 142: I think it is clearer to use `@var` for the cases where PHPStan analysis is deficient, rather than to add conditionals that will always succeed.

So until https://github.com/phpstan/phpstan/issues/9553 is implemented, I would do something like:

```php
/** @var \DOMNodeList $links */
$links = $xpath->query('//link');
foreach ($links as $element) {
```

We might even do `DOMNodeList<DOMElement>` and then the second `@var` should not be necessary.

tests/Unit/ItemTest.php
```diff
@@ -2834,7 +2834,7 @@ public function test_get_longitude(string $data, float $expected): void
     }
 
     /**
-     * @return array<array{string, string}>
+     * @return array<string, array<int, string|null>>
```
jtojnar, line 2837: We should keep the [tuple syntax](https://phpstan.org/writing-php-code/phpdoc-types#array-shapes):

```suggestion
     * @return array<string, array{string, string|null}>
```

tests/Unit/SimplePieTest.php
grimmdude: This test seems to be specifically checking that the content is the same as the control.  Would you like to use `assertInstanceOf` instead, or in addition to?

tests/Unit/RestrictionTest.php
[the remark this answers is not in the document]
grimmdude: Got it, maybe worth handling that in a separate branch?

tests/Unit/LocatorTest.php
grimmdude: Got it, done.

tests/Unit/ItemTest.php
grimmdude: Done ✅

src/Sanitize.php
```diff
@@ -688,11 +703,14 @@ protected function strip_tag(string $tag, DOMDocument $document, DOMXPath $xpath
...
                 for ($i = $number; $i > 0; $i--) {
-                    $child = $element->childNodes->item(0);
-                    $fragment->appendChild($child);
+                    if ($child = $element->childNodes->item(0)) {
+                        $fragment->appendChild($child);
+                    }
                 }
 
-                $element->parentNode->replaceChild($fragment, $element);
+                if ($parentNode = $element->parentNode) {
```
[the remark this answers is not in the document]
grimmdude: Done ✅

tests/Unit/SimplePieTest.php
jtojnar: In addition to. This would ensure that `get_item()` actually returns `Item`, and not `null`, so we would be sure we can call `get_content()` on it.

grimmdude: Got it, done ✅

grimmdude:
Friendly bump on this one, @jtojnar

src/Enclosure.php
```diff
@@ -992,7 +992,10 @@ public function embed($options = '', bool $native = false)
         // Flash Media Player file types.
         // Preferred handler for MP3 file types.
         elseif ($handler === 'fmedia' || ($handler === 'mp3' && $mediaplayer !== '')) {
-            $height += 20;
+            if (is_int($height)) {
+                $height += 20;
```
Alkarex, line 996: What is height is a float? Likely, given that it comes from a [`round()`](https://php.net/round) higher up.
I am afraid several of those changes might hide regressions

jtojnar: Looks like that code is already broken. If the default `$width === 'auto'` is used, it will have type [`int|float|'100%'`](https://github.com/simplepie/simplepie/blob/9385c54382edec0a768fdbe310c3e580c1b5dc81/src/Enclosure.php#L948). Later it is being [tested against `'auto'`](https://github.com/simplepie/simplepie/blob/9385c54382edec0a768fdbe310c3e580c1b5dc81/src/Enclosure.php#L956), which the variable cannot hold.

I guess this is another feature to deprecate, just the mention of flash makes me gag.

jtojnar (commented):
I pushed some more improvements to https://github.com/grimmdude/simplepie/pull/1 but am still reviewing this PR (currently 31/36 files).

src/Enclosure.php
grimmdude: I've gone ahead and adjusted these to use `is_numeric()` instead.

Art4:
The dependent PRs #892, #890 and #894 are merged.

src/Misc.php
```diff
@@ -96,7 +96,7 @@ public static function get_element(string $realname, string $string)
                         if (count($attribs[$j]) === 2) {
```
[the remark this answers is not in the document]
jtojnar: Switched to loop in #928

jtojnar (approved):
Thank you very much. With the latest changes, I consider the PR good enough to merge.

Not much option other than squashing everything together.

Art4 (changes requested):

Issue #642, opened 2020-03-22
how can we get image 


rosshhun:
I am wondering how can I get a particular image from the feed or an image from the feed.
and one more thing is there any official forum for simplepie.org?

skyzyx:
> I am wondering how can I get a particular image from the feed or an image from the feed.

Not enough information here. Are you talking about the `<channel><image>` element, or are you talking about some arbitrary image inside the HTML content of the `<item>`?

> and one more thing is there any official forum for simplepie.org?

Not for about 10 years now.

sayirosshhun:
talking about both the <channel><image> element and image inside the HTML content of the <item>

skyzyx:
Post the code to show what you've already tried, and those of us who have a few minutes might be able to help you tweak your code to get it working.

arena94:
i solved this a while ago by extending class SimplePie_Item with a get _image() function.

Looking for code sample ?  take a look at a wordpress plugin (dashpress)

https://wordpress.org/plugins/dashpress/

the extended class is in dashpress/dbp-admin/class/DBP_SimplePie_Item.class.php

Art4:
The `get_image()` and `get_images()` methods were build using `get_thumbnails()` and can be found here: https://plugins.trac.wordpress.org/browser/dashpress/tags/5.0/dbp-admin/class/DBP_SimplePie_Item.class.php#L46

@mblaney I think this can be closed.

marc-farre:
I've sent this PR: https://github.com/simplepie/simplepie/pull/858
I hope it helps!

PR #858, opened 2024-02-07
New method Item::get_images() for retrieving images from an item

marc-farre:
From https://github.com/simplepie/simplepie/issues/642#issuecomment-1401683178


jtojnar:
What is the goal of this? It appears to fetch whatever object fitting the image hole but how is that even useful? As a app, you generally treat enclosures and thumbnails differently as they have different properties.

Additionally, there are weird undocumented heuristics. I would like to move stuff like this out the SimplePie core since they just tend to bitrot.

Then there is the issue with license. Are you the original author and allowed to re-license it under `BSD-3-Clause` licence used by SimplePie? It appears that the project which this was taken from [uses WTFPL](https://plugins.trac.wordpress.org/browser/dashpress/trunk/license.txt), which is [famously controversial](https://en.wikipedia.org/wiki/WTFPL#Reception). From a brief reading, it should probably be compatible but I am not a lawyer.

marc-farre:
Thanks for your answer.
So let's close this PR, I'm missing time to do more, sorry.

PR #865, opened 2024-05-13
Fix typos

szepeviktor:
Found few misspellings.


szepeviktor:
For the willing contributor: `.typos.toml`
```toml
[files]
extend-exclude = [
    ".git/",
    "src/Parse/Date.php",
    "src/Misc.php",
    "demo/for_the_demo/sifr.js",
    "demo/for_the_demo/source_files/sIFR-r245/",
    "demo/for_the_demo/favicons/",
]
ignore-hidden = false

[default]
extend-ignore-re = [
    "return '<!DOCTYPE html.*'",
    "'curren;?' =>",
    "&#8220;Unio,",
]

[default.extend-identifiers]
# TODO
"datas" = "datas"
"nsIInterfaceRequestor" = "nsIInterfaceRequestor"
```

src/Misc.php
```diff
@@ -292,13 +292,13 @@ public static function change_encoding(string $data, string $input, string $outp
 
         // We fail to fail on non US-ASCII bytes
         if ($input === 'US-ASCII') {
-            static $non_ascii_octects = '';
-            if (!$non_ascii_octects) {
+            static $non_ascii_octets = '';
```
Art4, line 295: Renaming this property could be a breaking change.

Art4 (commented):
LGTM, but renaming `Misc::$non_ascii_octects` could be a breaking change. We have to check this first.

src/Misc.php
szepeviktor: Yes. I do not know how to handle this.

jtojnar: What do you mean? Is there some way to access the static variable outside the scope of this function?

I do not see any mention in the [docs](https://www.php.net/manual/en/language.variables.scope.php#example-283) and reusing the name in a child method seems to create independent variable: https://3v4l.org/hKPjc

jtojnar:
> .typos.toml

Is it this tool https://github.com/crate-ci/typos?

Might be nice to add it to CI.

src/Misc.php
szepeviktor: I mean I do not know to handle a breaking change.

szepeviktor:
> Is it this tool https://github.com/crate-ci/typos?

Yes.
https://github.com/crate-ci/typos/blob/master/docs/github-action.md

src/Misc.php
jtojnar: I meant that I do not see how this could be a breaking change.

Art4: > I meant that I do not see how this could be a breaking change.

You are right. I missed the assignment of the empty string `= '';`.

Art4 (approved):

jtojnar (approved):

src/Misc.php
Art4: This variable was introduced Dec 2007 with https://github.com/simplepie/simplepie/commit/a9633c8258d75331c00722fcea70688e0063832b

We should remove the `global` statement while we are renaming the variable.

jtojnar: Well, it is just an optimization to avoid creating the list multiple times. But we could just use hard-code it to `"\x80\x81\x82\x83\x84\x85\x86\x87\x88\x89\x8a\x8b\x8c\x8d\x8e\x8f\x90\x91\x92\x93\x94\x95\x96\x97\x98\x99\x9a\x9b\x9c\x9d\x9e\x9f\xa0\xa1\xa2\xa3\xa4\xa5\xa6\xa7\xa8\xa9\xaa\xab\xac\xad\xae\xaf\xb0\xb1\xb2\xb3\xb4\xb5\xb6\xb7\xb8\xb9\xba\xbb\xbc\xbd\xbe\xbf\xc0\xc1\xc2\xc3\xc4\xc5\xc6\xc7\xc8\xc9\xca\xcb\xcc\xcd\xce\xcf\xd0\xd1\xd2\xd3\xd4\xd5\xd6\xd7\xd8\xd9\xda\xdb\xdc\xdd\xde\xdf\xe0\xe1\xe2\xe3\xe4\xe5\xe6\xe7\xe8\xe9\xea\xeb\xec\xed\xee\xef\xf0\xf1\xf2\xf3\xf4\xf5\xf6\xf7\xf8\xf9\xfa\xfb\xfc\xfd\xfe\xff"`

But that does not have to go to this PR.

Art4: What do you think about a private static property in the `Misc` class.

jtojnar: I do not have a strong preference. On one hand, moving it to the class would allow us to use `const`. On the other hand, the value is only used by a single function so it makes sense to me for it to be defined locally.

jtojnar:
@mblaney could you please merge #862 and this PR?

PR #866, opened 2024-06-15
Fix SimplePie autodiscovery for text/xml HTML pages

Alkarex:
https://github.com/FreshRSS/FreshRSS/pull/1265
https://github.com/FreshRSS/FreshRSS/issues/1264


Alkarex:
Sorry, wrong manipulation

PR #868, opened 2024-06-18
PHPDoc avoid aliases for scalar types

Alkarex:
https://php.net/language.types.declarations
image

Also helps PHPStan, which otherwise makes another lookup as it might be in the SimplePie namespace, and which fails in some cases.


Alkarex:
If you merge https://github.com/simplepie/simplepie/pull/869 first, the PHPStan tests will pass

jtojnar (approved):
LGTM, thanks.

Art4 (approved):

PR #869, opened 2024-06-18
Fix PHPStan for PHP 7.3

Alkarex:
The signature of `fread()` has changed from PHP 7.3 to 7.4

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -145,7 +145,7 @@ class Parser implements RegistryAware
                 do {
                     $stream_data = fread($stream, 1048576);
                     // NB: At some point between PHP 7.3 and 7.4, the documented signature for `fread()` has changed
-                    // from returning `string` to returning `string|false`, and in older versions, it used to be `string|null`,
+                    // from returning `string` to returning `string|false`,
                     // hence the falsy check:
                     if (!xml_parse($xml, $stream_data == false ? '' : $stream_data, feof($stream))) {
                         $this->error_code = xml_get_error_code($xml);
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -144,7 +144,7 @@ class Parser implements RegistryAware
                 //Parse by chunks not to use too much memory
                 do {
                     $stream_data = fread($stream, 1048576);
-                    // NB: At some point between PHP 7.3 and 7.4, the documented signature for `fread()` has changed
+                    // NB: At some point between PHP 7.3 and 7.4, the signature for `fread()` has changed
                     // from returning `string` to returning `string|false`, hence the falsy check:
                     if (!xml_parse($xml, $stream_data == false ? '' : $stream_data, feof($stream))) {
                         $this->error_code = xml_get_error_code($xml);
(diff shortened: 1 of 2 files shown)


src/Parser.php
```diff
@@ -144,7 +144,7 @@ public function parse(string &$data, string $encoding, string $url = '')
                 //Parse by chunks not to use too much memory
                 do {
                     $stream_data = fread($stream, 1048576);
-                    if (!xml_parse($xml, $stream_data === false ? '' : $stream_data, feof($stream))) {
```
jtojnar, line 147: That sounds like a PHPStan bug. PHP 7.3 returns false on failure too: https://3v4l.org/SfHsD

Maybe add a comment or ignore this specific error?

Alkarex: At some point around PHP ~7.4, the documented signature for `fread()` has changed from returning `string` to returning `string|false`.
Anyway, if `$stream_data` is falsy, it does not make sense to attempt an XML parsing

Alkarex: Comments added: https://github.com/simplepie/simplepie/pull/869/commits/3bf07340e5dcb43ffe1eb19311f4708aa8e37c81

jtojnar: AFAICS, it was never claiming `null`, [not even in PHP 3](https://github.com/php/doc-en/blob/ec1ea24811105ebafb02b842c98195186f24cfa1/functions/filesystem.sgml#L947). It did appear to return it [prior to 4.3.5](https://3v4l.org/SfHsD) patch release but that’s probably not even worth a footnote. Even before PHP 7.0, examples were using strict comparison to `false`: https://github.com/php/doc-en/commit/877e82bf2fbd1c0736f078035cc40ddb71fa31aa

But I agree that this should be safe, if ugly. I believe the only possible values that match the “real” type signature and [`== false`](https://www.php.net/manual/en/language.types.boolean.php#language.types.boolean.casting) are:

- `''` – the ternary will not change it
- `'0'` – I guess this can only occur for pathological XML files, where the last chunk consists of a single `0` (e.g. `'<a>' . str_repeat('f', 1048576 - strlen('<a></a>')) . '</a>0'`). This change will mask an error in this case but I would consider the failure recoverable so the new behaviour is actually better, if unexpected.
- `false` – this can only happen on stream failure

Alkarex: Comment simplified https://github.com/simplepie/simplepie/pull/869/commits/215365a3980b58e33cd66008770ca6f6a30bd6a3

jtojnar: Apparently, https://github.com/phpstan/phpstan-src/pull/3105 is responsible.

Alkarex: Great job for precise test 👍🏻
In any case, I think the current PR would make the behaviour more robust AND pass with PHPStan

Alkarex: https://github.com/simplepie/simplepie/pull/869/commits/55578b4373b26238e942d6f78dfad82057fb1a1b

Alkarex: While you were at it, did you check the `feof()` behaviour?

jtojnar: Yes, `feof` appears to return `true` after unmounting on all tested versions. So the infinite loop is probably not an issue. I only managed to get `feof` and `fread` both return `false` when I `fclose`d the handle (which probably falls under bogus resource mentioned in https://github.com/phpstan/phpstan-src/pull/3105).

jtojnar (approved):

Art4 (approved):

PR #871, opened 2024-07-22
Extract long repeating sanitization bases into variable

jtojnar:
The following regular expression matches:

    ^( +)(\$[a-z_]+) = \$this->sanitize\((.+)\['data'\](, .+), \$this->get_base\(\3\)

were replaced by:

    \1\2 = \3;\n\1\2 = \$this->sanitize\(\2\['data'\]\4, \$this->get_base\(\2\)

### DIFF
--- a/src/Item.php
+++ b/src/Item.php
@@ -474,7 +474,8 @@ class Item implements RegistryAware
                 $name = $this->sanitize($contributor['child'][\SimplePie\SimplePie::NAMESPACE_ATOM_10]['name'][0]['data'], \SimplePie\SimplePie::CONSTRUCT_TEXT);
             }
             if (isset($contributor['child'][\SimplePie\SimplePie::NAMESPACE_ATOM_10]['uri'][0]['data'])) {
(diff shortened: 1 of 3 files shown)


Art4 (approved):

src/Item.php
```diff
@@ -474,7 +474,8 @@ public function get_contributors()
                 $name = $this->sanitize($contributor['child'][\SimplePie\SimplePie::NAMESPACE_ATOM_10]['name'][0]['data'], \SimplePie\SimplePie::CONSTRUCT_TEXT);
             }
             if (isset($contributor['child'][\SimplePie\SimplePie::NAMESPACE_ATOM_10]['uri'][0]['data'])) {
```
jtojnar, line 476: Actually, I wonder if we should not just replace the `isset`s with

```suggestion
            if (($uri = $contributor['child'][\SimplePie\SimplePie::NAMESPACE_ATOM_10]['uri'][0] ?? null) !== null && isset(['data'])) {
```

It is ugly but I think it is still easier to read than the current repeating long chains.

mblaney: this pr looks good @jtojnar, up to you if you want to add more cleanup

jtojnar: Yeah, we can do more cleanup separately.

PR #876, opened 2024-08-27
ci: Add PHP 8.4

jtojnar:
PHP 8.4 is in beta, with final version scheduled for November so it is time to start testing it.

Add it to CI matrix, allowing failures for now. Do not allow failures for PHP 8.3 since that has been stable for ages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,10 +44,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ['7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3']
+        php:
+          - '7.2'
+          - '7.3'
+          - '7.4'
+          - '8.0'
+          - '8.1'
+          - '8.2'
+          - '8.3'
+          - '8.4'
 
     name: "PHP: ${{ matrix.php }}"
-    continue-on-error: ${{ matrix.php == '8.3' }}
+    continue-on-error: ${{ matrix.php == '8.4' }}
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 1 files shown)


jrfnl (commented):
Came here to do the same. Thanks for creating the PR.

The conditions on (now) line 75 and 79 also need changing (when to run `composer install with `--ignore-platform-reqs`).

I'd also like to suggest running the `test-compiled` job against `php: ['7.2', 'latest']`, i.e. high/low of the supported PHP versions.
`setup-php` will automatically update `'latest'` to a new PHP version once released.

jrfnl:
Oh and would be great if this change would also be backported to the 1.8 branch.

jrfnl:
Oh and one more thing - what with the changes in PHPUnit 10 and the test builds on PHP 8.2 - 8.4 running on PHPUnit 10, the build will currently **not** fail on PHP deprecations/notices/warnings. It will not even show these either...

This may be an intentional choice from the project, but if it is not, I'd strongly recommend updating the configuration/CI.

jrfnl:
@jtojnar Now the other PHP 8.4 (+ the PHPStan) PRs have been merged, could you rebase the PR and fix up line 66/67 + 70/71 per my feedback ? I believe the build should now pass without any deprecation notices.

> what with the changes in PHPUnit 10 and the test builds on PHP 8.2 - 8.4 running on PHPUnit 10, the build will currently not fail on PHP deprecations/notices/warnings. It will not even show these either...
>
> This may be an intentional choice from the project, but if it is not, I'd strongly recommend updating the configuration/CI.

Do you want to tackle this with your new maintainer status (congrats!) or would you like me to propose a change to get this working properly ?

jrfnl:
Oh and it would be nice for this PR to be backported to the 1.8 branch too...

jtojnar:
Thanks. Will try to apply your suggestions soon.

> > what with the changes in PHPUnit 10 and the test builds on PHP 8.2 - 8.4 running on PHPUnit 10, the build will currently not fail on PHP deprecations/notices/warnings. It will not even show these either...
>
> > This may be an intentional choice from the project, but if it is not, I'd strongly recommend updating the configuration/CI.
>
> Do you want to tackle this with your new maintainer status (congrats!) or would you like me to propose a change to get this working properly ?

Would be great if you can do this. I am currently trying to chew through the PR backlog.

jtojnar:
I have updated the PR except for adding oldest/latest to `test-compiled` as the matrix is already pretty big (saving the environment a bit :stuck_out_tongue_winking_eye:) and I want to remove it soon anyway.

jrfnl (approved):

Art4 (approved):

PR #878, opened 2024-09-08
Fix CI for PHPStan 1.12

Alkarex:
CI was broken due to not pinning `composer.lock` AND allowing updates at `1.x.x` level, which may contain breaking changes, e.g. for PHPStan.

```
------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   src/IRI.php
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  :304   Offset 3 on array{0: string, 1: '', scheme: null, 2: string, 3: string, authority: string, 4: string, path: string, ...}|array{0: string, 1: non-empty-string, scheme: string, 2: string, 3: string, authority: string, 4: string, path: string, ...} in isset()
         always exists and is not nullable.
  :307   Offset 5 on array{0: string, 1: '', scheme: null, 2: string, 3: string, authority: string|null, 4: string, path: string, ...}|array{0: string, 1: non-empty-string, scheme: string, 2: string, 3: string, authority: string|null, 4: string, path: string, ...}
         in isset() always exists and is not nullable.
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.19 || ^3.8",
         "mf2/mf2": "^0.5.0",
-        "phpstan/phpstan": "^1.11",
+        "phpstan/phpstan": "~1.12.2",
         "phpunit/phpunit": "^8 || ^9 || ^10",
         "psr/http-client": "^1.0",
         "psr/http-factory": "^1.0",
--- a/src/IRI.php
+++ b/src/IRI.php
@@ -301,12 +301,10 @@ class IRI
             if ($match[1] === '') {
                 $match['scheme'] = null;
             }
-            if (!isset($match[3]) || $match[3] === '') {
+            if ($match[3] === '') {
                 $match['authority'] = null;
             }
-            if (!isset($match[5])) {
-                $match['path'] = '';
-            }
+            $match['path'] = '';
             if (!isset($match[6]) || $match[6] === '') {
                 $match['query'] = null;
             }
--- a/src/IRI.php
+++ b/src/IRI.php
@@ -304,7 +304,6 @@ class IRI
             if ($match[3] === '') {
                 $match['authority'] = null;
             }
-            $match['path'] = '';
             if (!isset($match[6]) || $match[6] === '') {
                 $match['query'] = null;
             }


Alkarex:
The CI approach of this repo is broken.
I can gladly make another PR pinning `composer.lock` if there is such a wish from the maintainers.
It is not nice to have to fix unrelated parts of the code when making a PR.
For instance now, I am fixing PHPStan, and PHPUnit is also failing...

src/IRI.php
```diff
@@ -301,12 +301,9 @@ protected function parse_iri(string $iri)
             if ($match[1] === '') {
                 $match['scheme'] = null;
             }
-            if (!isset($match[3]) || $match[3] === '') {
+            if ($match[3] === '') {
                 $match['authority'] = null;
             }
-            if (!isset($match[5])) {
```
jtojnar, line 307: Looks good, `$match[3]` will never be unset since `$match[5]` is matched unconditionally:
https://3v4l.org/vifvK

composer.json
```diff
@@ -32,7 +32,7 @@
     "require-dev": {
         "friendsofphp/php-cs-fixer": "^2.19 || ^3.8",
         "mf2/mf2": "^0.5.0",
-        "phpstan/phpstan": "^1.11",
+        "phpstan/phpstan": "~1.12.2",
```
jtojnar, line 35: Even patch versions can sometimes introduce breakage but [pinning the minor version](https://getcomposer.org/doc/articles/versions.md#tilde-version-range-) will reduce the change a bit.

jtojnar, line 35: We will want to update it regularly but I guess we can do that later. Not sure how well [dependabot](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#versioning-strategy) works with tilde constraint, though.

jtojnar, line 35: Unfortunately, we cannot use single `composer.lock` since we support wide range of PHP versions that require different `php-cs-fixer` versions.

jtojnar (approved):

src/IRI.php
```diff
@@ -301,12 +301,10 @@ protected function parse_iri(string $iri)
             if ($match[1] === '') {
```
jtojnar, line 301: We could simplify this further by passing [PREG_UNMATCHED_AS_NULL](https://www.php.net/manual/en/pcre.constants.php#constant.preg-unmatched-as-null) to `preg_match` (available since PHP 7.2).

composer.json
Alkarex: Does it makes sense at all to run `php-cs-fixer` with multiple PHP versions? There is a single code base / syntax.

Alkarex: FWIW, we have been using PHPCS + PHPCBF on https://github.com/FreshRSS/FreshRSS/ with multiple PHP versions for years without such problems

jtojnar: The issue is that Composer will always try to install all development dependencies and installation will fail for incompatible PHP versions.

Alkarex:
Anything blocking here? 🐢

jtojnar:
@mblaney Can you please merge this, https://github.com/simplepie/simplepie/pull/881 and https://github.com/simplepie/simplepie/pull/880?

Art4 (approved):

Alkarex:
https://github.com/simplepie/simplepie/issues/884

PR #879, opened 2024-09-08
Add phpcs.xml

Alkarex:
(I was tired of seeing red errors everywhere when opening SimplePie)
image

The current `.php-cs-fixer.dist.php` does not seem to do much when it comes to live feedback, so consider adding a standard configuration file for PHPCS so that it works with the correct code style convention in IDEs
https://github.com/PHPCSStandards/PHP_CodeSniffer
Example of extension for VSCode https://marketplace.visualstudio.com/items?itemName=ValeryanM.vscode-phpsab

With that config, there are still many PSR12 errors, but at least most are gone. By the way, `phpcbf` could automatically fix many of those (for instance missing spaces in string concatenations `$a.'b'`). I can gladly send a PR if there is interest.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -26,3 +26,6 @@ indent_style = space
 [*.{yaml,yml}]
 indent_size = 2
 indent_style = space
+
+[*.xml]
+indent_style = tab
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ruleset name="SimplePie">
+	<arg name="extensions" value="php"/>
+	<exclude-pattern>./.git/</exclude-pattern>
+	<exclude-pattern>./compatibility_test/</exclude-pattern>
+	<exclude-pattern>./vendor/</exclude-pattern>
+	<rule ref="PSR12">
+		<exclude name="PSR1.Files.SideEffects.FoundWithSymbols"/>
+		<exclude name="PSR1.Methods.CamelCapsMethodName.NotCamelCaps"/>
+		<exclude name="PSR12.ControlStructures.ControlStructureSpacing.FirstExpressionLine"/>
+		<exclude name="Squiz.ControlStructures.ControlSignature.SpaceAfterCloseBrace"/>
+	</rule>
+	<rule ref="Generic.Files.LineLength">
+		<properties>
+			<property name="lineLimit" value="165"/>
+			<property name="absoluteLineLimit" value="400"/>
+		</properties>
+	</rule>
+</ruleset>


Alkarex:
P.S. Merge https://github.com/simplepie/simplepie/pull/878 because the CI is broken...

jtojnar:
We already use PHP-CS-Fixer so it does not make much sense to also introduce an PHP_CodeSniffer, IMO.

- Doing so would require us to keep ensuring its config does not conflict with PHP-CS-Fixer.
- There appears to be a [VSCode extension](https://github.com/junstyle/vscode-php-cs-fixer) for people who want that. But generally, I would recommend just running `composer fix` before every commit or setting the editor to run the command on every file save.

> With that config, there are still many PSR12 errors, but at least most are gone. By the way, `phpcbf` could automatically fix many of those (for instance missing spaces in string concatenations `$a.'b'`). I can gladly send a PR if there is interest.

We can enable that in PHP-CS-Fixer with [`concat_space`](https://cs.symfony.com/doc/rules/operator/concat_space.html) rule. Apparently, it was not included in PSR-12 ruleset initially due to the [rule limitations](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/4943#discussion_r461724662) at the time. Or we can switch to [PER-CS2.0 ruleset](https://cs.symfony.com/doc/ruleSets/PER-CS2.0.html).

Alkarex:
The current PHP-CS-Fixer seems to be quite limited:
* It does not catch as many rules as the [main PHPCS](https://github.com/PHPCSStandards/PHP_CodeSniffer). Here is another example:
image
* It seems to be picky with PHP versions, while PHP_CodeSniffer works with all PHP versions supported by SimplePie (allowing pinning `composer.lock`)
* It does not offer modern IDE integration:
    * Live in-code warnings (see screenshot above)
    * Integration if the file explorer
image
    * Integration if the list of problems, allowing an overview + click to jump to the corresponding file at the right line
image
    * Code autoformat using the IDE default command(s) (which is the same for all language, all projects)
    * Code suggestions from the IDE using the correct format
* It does not seem to allow having different rules in the same source tree, for instance when SimplePie is in a sub-folder of a larger project.

See also the difference in popularity:
image
vs.
image

jtojnar:
If we wanted to switch to PHP_CodeSniffer, we should drop PHP-CS-Fixer to avoid the conflicts.

Though, I am not convinced it is overall better.

> * It does not catch as many rules as the [main PHPCS](https://github.com/PHPCSStandards/PHP_CodeSniffer).

This particular issue should be fixed by https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/7901. Unfortunately, it is stuck in review purgatory for now.

Though, both of the linters have some blind spots. IIRC, last time I tried to use PHP_CodeSniffer, it did not handle some newer PHP constructs well. (Though that would not be an issue for SimplePie which still supports PHP 7.2.)

See also https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/3459#issuecomment-359234259 on their differences.

> * It seems to be picky with PHP versions, while PHP_CodeSniffer works with all PHP versions supported by SimplePie (allowing pinning `composer.lock`)

Agreed, this would be the main benefit IMO.

> * It does not offer modern IDE integration:
>   
>   * Live in-code warnings (see screenshot above)
>   * Integration with the file explorer
>   * Integration with the list of problems, allowing an overview + click to jump to the corresponding file at the right line

You should not need in-code warnings. Unlike with PHP_CodeSniffer, all errors reported by php-cs-fixer are also fixable by it so you can just have the editor format-on-save and you should never see any errors.

>   * Code autoformat using the IDE default command(s) (which is the same for all languages, all projects)

I see that listed as supported in [usage](https://github.com/junstyle/vscode-php-cs-fixer?tab=readme-ov-file#usage).

>   * Code suggestions from the IDE using the correct format

This would be nice to have but formatting on save should be sufficient and much easier to implement.

> * It does not seem to allow having different rules in the same source tree, for instance when SimplePie is in a sub-folder of a larger project.

I do not see that as a significant deciding factor.

Projects that vendor dependencies often add the vendored directories to the formatter’s ignore list since formatting makes updating harder. But if you really want, you can just invoke the formatter separately on the vendored tree.

> See also the difference in popularity:

The first result on [DuckDuckGo](https://duckduckgo.com/?q=php-cs-fixer+vs+code&t=ffab&ia=web) is more than [four times more popular](https://marketplace.visualstudio.com/items?itemName=junstyle.php-cs-fixer) than the PHPCS extension.

The https://github.com/PHP-CS-Fixer/PHP-CS-Fixer repo also has four times the stars than https://github.com/PHPCSStandards/PHP_CodeSniffer

Alkarex:
A bunch of the rules supported by PHP_CodeSniffer are automatically fixable by phpcbf, although some do indeed require human intervention (e.g. writing mandatory comments, adding a namespace, shortening lines, etc.)

Art4:
@jtojnar 

> Or we can switch to [PER-CS2.0 ruleset](https://cs.symfony.com/doc/ruleSets/PER-CS2.0.html).

👍 

@Alkarex 

> (I was tired of seeing red errors everywhere when opening SimplePie)

What about using the `phpcs.xml` only on your machine?

```bash
echo 'phpcs.xml' > .git/info/exclude
```

Alkarex:
> What about using the `phpcs.xml` only on your machine?

It is in my [own fork of FreshRSS](https://github.com/FreshRSS/simplepie) so I have solutions. It is more that when navigating between multiple PHP projects, it is nice that it works more or less out of the box when checking the repo out in one's favourite IDE, which is currently not the case for SimplePie. Similar reasoning than `.editorconfig`, which makes sure that it instantly works without having to *just add..., just run..., just change...* :-)

Alkarex:
I do not want to push this suggestion too much, especially since there are other more pressing PRs waiting, but one last clarification:

> You should not need in-code warnings. Unlike with PHP_CodeSniffer, all errors reported by php-cs-fixer are also fixable by it so you can just have the editor format-on-save and you should never see any errors.

That is maybe the main misunderstanding: when opening the SimplePie repo, an IDE with PHP support will report many warnings / errors, because there is no declaration of the conventions used in that repo, which are not fully PSR-12 (but that would not matter with such a declaration).

jtojnar:
> That is maybe the main misunderstanding: when opening the SimplePie repo, an IDE with PHP support will report many warnings / errors, because there is no declaration of the conventions used in that repo, which are not fully PSR-12 (but that would not matter with such a declaration).

If the fixes are compatible with what PHP-CS-Fixer expects, it should be fine to submit them.

Alkarex:
> If the fixes are compatible with what PHP-CS-Fixer expects, it should be fine to submit them.

By default (without `phpcs.xml`), there are plenty of errors, which are incompatible with the conventions used by PHP-CS-Fixer in this project. For instance, SimplePie uses camel caps, which is not standard. Hence the value of committing a declaration of the conventions used in the project. I can make `phpcs.xml` to comply with the desired SimplePie conventions.

But I can happily make an autofix of the several types of problems not supported by PHP-CS-Fixer (dot spacing, indents in several situations, etc.).

jtojnar:
> > * It seems to be picky with PHP versions, while PHP_CodeSniffer works with all PHP versions supported by SimplePie (allowing pinning `composer.lock`)
> 
> Agreed, this would be the main benefit IMO.

Also, we do have other dependencies with different major version per PHP version like `phpunit` and `psr/simple-cache`. For the former, we could at least use a proxy `symfony/phpunit-bridge` but nothing for the latter. So this point is moot.

> By default (without `phpcs.xml`), there are plenty of errors, which are incompatible with the conventions used by PHP-CS-Fixer in this project. For instance, SimplePie uses camel caps, which is not standard. Hence the value of committing a declaration of the conventions used in the project.

Why does it assume the coding style when there is no `phpcs.xml`?

> I can make `phpcs.xml` to comply with the desired SimplePie conventions.

Well, the issue would be that unless we disable everything in `phpcs.xml`, the configs will almost certainly conflict eventually. Then people with PHP_CodeSniffer will get output that is rejected by CI yet valid according to the config file.

> But I can happily make an autofix of the several types of problems not supported by PHP-CS-Fixer (dot spacing, indents in several situations, etc.).

:+1:

Alkarex:
> Why does it assume the coding style

Because this is what IDEs with support for a language do?

Anyway, spending too much time here. There are other more important PRs, which have waited for ages

PR #880, opened 2024-09-12
PHP 8.4 | Fix implicitly nullable parameters

jrfnl:
PHP 8.4 deprecates implicitly nullable parameters, i.e. typed parameter with a `null` default value, which are not explicitly declared as nullable.

Includes updating the documentation to match.

Ref: https://wiki.php.net/rfc/deprecate-implicitly-nullable-types


jrfnl:
Failing build is unrelated to this PR and should be fixed via #878

jtojnar (approved):
LGTM, thanks.

Alkarex (approved):

Art4 (approved):

Alkarex:
https://github.com/simplepie/simplepie/issues/884

Alkarex:
@jtojnar A little detail about merging: I can see you have forced-pushed, which destroys the git history, comments, etc. and also remove the ownership of the commit from the original author (@jrfnl ). I can personally better like to use the GitHub flow of updating the branch (just click on the button) and then *squash and merge* to have a nice git history in the main branch.

Art4:
:+1: for not using force-pushes.

jtojnar:
I did not actually force push, I used the “Update branch” button, choosing the “Update branch with rebase” option for cleaner git history.

I am not sure what you mean by destroying the history – squashing would do that but rebase preserves the original commit author as well as the commit messages. The only thing that changes is the committer tag, which makes sense as I am integrating the commit into the repository.

Alkarex:
Ah, right, *rebase*  is the culprit. It is not needed if you use *Squash and merge* just after.
Well, all the actions which rewrite history break things such as original names, signatures, associated comments, links, etc.

jrfnl:
As the author of these commits - I'm generally not a fan of squash-merging when the original commits are atomic commits with detailed commit messages - like mine are, as it destroys the context (commit + message being closely bound).

When there is one atomic commits and the follow-up commits are "fix up" commits, that's, of course, a different matter.

jrfnl:
Oh and as a side-note, I'll happily rebase any PR of mine whenever you think it's needed.

Alkarex:
There are different preferences of course. I typically avoid all the processes rewriting the git history, including rebase. It breaks `git pull` (especially problematic when someone is testing a combination of multiple PRs), breaks the comments in GitHub, breaks the links, breaks the "see changes since last time" (especially useful for the reviews), etc.

*Squash and merge* also allows a combined message for the whole PR, while keeping the PR for the details of how it ended up as it is (which is lost with rebase). Without it, it gave the following in the `master` branch, which is not so clean:

image

But those things are details, personal preferences, and I am happy to see the PRs merged faster :-)

jtojnar:
> Ah, right, _rebase_ is the culprit. It is not needed if you use _Squash and merge_ just after.
> Well, all the actions which rewrite history break things such as original names, signatures, associated comments, links, etc.

Well, squash merge will suffer from exactly the same problems, in addition to removing the context as @jrfnl mentions.

Yes, the context will remain in the PR itself but it is pretty much useless there, as tools like `git blame` will not be able to access it. Also GitHub might not be here forever.

So I would argue that all the relevant context (especially the reasoning behind any non-trivial decisions) should be included in the commit messages or code itself whenever possible. 

Atomic commits also help since when there is only one thing the commit message can refer to, the context is clearer.

> There are different preferences of course. I typically avoid all the processes rewriting the git history, including rebase. It breaks `git pull` (especially problematic when someone is testing a combination of multiple PRs), breaks the comments in GitHub, breaks the links, breaks the "see changes since last time" (especially useful for the reviews), etc.

I agree that history rewrites are problematic and should be avoided on persistent public branches. But I generally consider PR branches ephemeral where everything goes. Especially before the merge itself, I think rebase/squash makes sense since it will actually result in the PR containing the same commits as the target branch (unlike the default Merges other than the fast-forward one). And then the branch is typically deleted anyway.

It is true that history rewrites make stacking multiple PRs together annoying with plain git but nowadays, there are tools that make this a breeze like one of the following:

- https://github.com/gitbutlerapp/gitbutler
- https://book.git-ps.sh/patch-stack-workflow
- https://github.com/gitext-rs/git-stack
- https://engineering.fb.com/2022/11/15/open-source/sapling-source-control-scalable/
- https://stacked-git.github.io/
- https://timothya.com/blog/git-stack/

Looks like there are new ones every time I look, I should really start using one of them :joy_cat: 

> _Squash and merge_ also allows a combined message for the whole PR, while keeping the PR for the details of how it ended up as it is (which is lost with rebase). Without it, it gave the following in the `master` branch, which is not so clean:

Right, I do not trust GitHub with the context. What is not accessible through `git blame` is annoying to dig up – I have already spent hours blaming SimplePie’s source code and often, the only context was a link to a bug tracker or forum that was lost to time.

If the commits are atomic and contain all the context like the @jrfnl’s ones from #881, there should be no need to ever go to the PR. And if you for some reason need it, you can still access the PR from the commit in GitHub UI.

And IMO, it is still cleaner than this :wink: 

Railroad mess of commits

> But those things are details, personal preferences, and I am happy to see the PRs merged faster :-)

Right. It would be good to settle on a common policy, though.

jrfnl:
> Yes, the context will remain in the PR itself but it is pretty much useless there, as tools like `git blame` will not be able to access it. Also GitHub might not be here forever.

Hear hear! It's something I always try to mention in my git talks too and why I try to always put all the relevant info in the commit message.
PRs and issues can be lost in time when platforms come and go (think: Sourceforge, PEAR etc). Commit messages are here to stay and transfer with the code to whatever the next platform is.

Alkarex:
I do not really mind the policies here, but just to clarify a few things:

> when the original commits are atomic commits

That is a big assumption to make, which requires the original author to be around, and makes review processes more cumbersome.

Furthermore, if a PR contains multiple unrelated commits, it should be multiple PRs instead

> like mine are

When the original author happens to provides a PR that is better suited for a merged commit, and if not change is needed from the maintainers' side (neither in the code nor in the commit messages - which might be rare I believe), then no reason for a squash indeed

> When there is one atomic commits and the follow-up commits are "fix up" commits, that's, of course, a different matter.

Indeed, and it is mostly this scenario (which I believe to be quite usual), which I am discussing here

> as it destroys the context (commit + message being closely bound)

The squash commit message can be anything the maintainer / merger wants including a better message than what the original author(s) may have provided or not. Especially true at the end of a review process, when the explanation to be put inside the message may have changed.
By default, in the case of a PR with a single commit, the message is the same as the commit.
It is though true that by default, the commit message is a list of the titles of the different commits of the PR, which might not be the best possible message.

> squash merge will suffer from exactly the same problems

We were not talking about the same situation. My point is that `rebase` and other forms of git history rewriting and forced pushed _inside a PR_ break the history of that PR and a number of other things such as comments, help to review, information about multiple authors, etc.

> tools like git blame will not be able to access it

`git blame` works perfectly fine with squashed commits, and even better than with merged commits (not to be confused with commit message aspects as explained just above)

> I would argue that all the relevant context (especially the reasoning behind any non-trivial decisions) should be included in the commit messages or code itself whenever possible

That is a good principle but independent on whether squashed commits are used or not

> What is not accessible through git blame is annoying to dig up

Indeed, good principle, but orthogonal to using squashed commits or not, and actually this can be made even *better*  thanks to squashed commits, as the maintainer has full control about the commit message

> If the commits are atomic and contain all the context like the @jrfnl’s ones

Indeed, in this case no need for a squashed commit, but I do not believe this is the majority of the situations, and insisting on it would add quite some overhead

I am out :-)

PR #882, opened 2024-09-14
Gzdecode: Deprecate our gzip implementation

jtojnar:
PHP supports it natively since 5.4.

acaaa76ed6ca095cd6cf236cc1bef03cc3a41578 also started to use it, presumably to fix sites that were incorrectly declaring `deflate` method when they were using `gzip`.

And since we require PHP 7.2, and the zlib extension is already mandatory for the `gzinflate` and `gzuncompress` functions, let’s also remove the `funcion_exists` check.

### DIFF
--- a/src/Gzdecode.php
+++ b/src/Gzdecode.php
@@ -11,6 +11,8 @@ namespace SimplePie;
  * Decode 'gzip' encoded HTTP data
  *
  * @link http://www.gzip.org/format.txt
+ * @link https://www.php.net/manual/en/function.gzdecode.php
+ * @deprecated since SimplePie 1.9.0, use `gzdecode` function instead.
  */
 class Gzdecode
 {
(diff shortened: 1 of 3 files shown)


src/Gzdecode.php
```diff
@@ -11,6 +11,8 @@
  * Decode 'gzip' encoded HTTP data
  *
  * @link http://www.gzip.org/format.txt
+ * @link https://www.php.net/manual/en/function.gzdecode.php
+ * @deprecated Use `gzdecode` function instead.
```
Art4, line 15:
```suggestion
 * @deprecated since SimplePie 1.9.0, use `gzdecode` function instead.
```

Art4 (changes requested):
In the CHANGELOG.md we should also add a note in the `Changed` section that now `gzdecode()` instead of `SimplePie\Gzdecode` is used.

Art4 (approved):

PR #883, opened 2024-09-14
Fix cache_expiration_time

Alkarex:
Fix cache handling logic and associated tests. 
fix https://github.com/simplepie/simplepie/issues/830 ([BUG] Cache constantly updating).
Note: the default cache implementation (legacy file cache) is still broken for allowing HTTP 304.
Replaces https://github.com/simplepie/simplepie/pull/846

### DIFF
--- a/src/SimplePie.php
+++ b/src/SimplePie.php
@@ -1937,6 +1937,7 @@ class SimplePie
                             // Set raw_data to false here too, to signify that the cache
                             // is still valid.
                             $this->raw_data = false;
+                            $this->data['cache_expiration_time'] = $this->cache_duration + time();
                             $cache->set_data($cacheKey, $this->data, $this->cache_duration);
 
                             return true;
--- a/src/SimplePie.php
+++ b/src/SimplePie.php
@@ -1901,7 +1901,7 @@ class SimplePie
                     $this->data = [];
                 }
                 // Check if the cache has been updated
-                elseif (isset($this->data['cache_expiration_time']) && $this->data['cache_expiration_time'] > time()) {
+                elseif (empty($this->data['cache_expiration_time']) || $this->data['cache_expiration_time'] < time()) {
                     // Want to know if we tried to send last-modified and/or etag headers
                     // when requesting this file. (Note that it's up to the file to
                     // support this, but we don't always send the headers either.)
@@ -1925,6 +1925,7 @@ class SimplePie
                             $this->status_code = 0;
 
                             if ($this->force_cache_fallback) {
+                                $this->data['cache_expiration_time'] = $this->cache_duration + time();
                                 $cache->set_data($cacheKey, $this->data, $this->cache_duration);
 
                                 return true;
--- a/src/SimplePie.php
+++ b/src/SimplePie.php
@@ -1900,7 +1900,7 @@ class SimplePie
                     $this->data = [];
                 }
                 // Check if the cache has been updated
-                elseif (empty($this->data['cache_expiration_time']) || $this->data['cache_expiration_time'] < time()) {
+                elseif (!isset($this->data['cache_expiration_time']) || $this->data['cache_expiration_time'] < time()) {
                     // Want to know if we tried to send last-modified and/or etag headers
                     // when requesting this file. (Note that it's up to the file to
                     // support this, but we don't always send the headers either.)
--- a/tests/Integration/CachingTest.php
+++ b/tests/Integration/CachingTest.php
@@ -120,7 +120,7 @@ class CachingTest extends TestCase
      */
     public static function provideSavedCacheData(): array
     {
-        $defaultMtime = time();
+        $defaultMtime = time() - 1; // -1 to account for tests running within the same second
         $defaultExpirationTime = $defaultMtime + 3600;
 
         $expectDefaultDataWritten = [
@@ -246,10 +246,10 @@ class CachingTest extends TestCase
             [CacheInterface::class, $currentlyCachedDataWithNonFeedUrl,     $expectDataWithNewFeedUrl, $defaultMtime],
             // Check if the cache has been updated
             [Base::class,           $currentlyCachedDataIsUpdated,          $expectDefaultDataWritten, $defaultMtime],
-            [CacheInterface::class, $currentlyCachedDataIsUpdated,          $expectDefaultDataWritten, $defaultMtime],
+            [CacheInterface::class, $currentlyCachedDataIsUpdated,          $expectNoDataWritten,      $defaultMtime],
             // If the cache is still valid, just return true
             [Base::class,           $currentlyCachedDataIsValid,            $expectDefaultDataWritten, $defaultMtime],
-            [CacheInterface::class, $currentlyCachedDataIsValid,            $expectNoDataWritten,      $defaultMtime],
+            [CacheInterface::class, $currentlyCachedDataIsValid,            $expectDefaultDataWritten, $defaultMtime],
         ];
     }
 }


Alkarex:
Requires https://github.com/simplepie/simplepie/pull/878 first to pass the CI

src/SimplePie.php
```diff
@@ -1924,6 +1924,7 @@ protected function fetch_data(&$cache)
                             $this->status_code = 0;
 
                             if ($this->force_cache_fallback) {
+                                $this->data['cache_expiration_time'] = $this->cache_duration + time();
                                 $cache->set_data($cacheKey, $this->data, $this->cache_duration);
```
jtojnar, line 1928: If I understand this correctly, this just extends the cache lifetime (`$this->data` remains the same). Not sure why bb38d740c12a953dce4e2722a8d26d7a3ed2be3f made SimplePie store `cache_expiration_time` in the cache when `set_data` already takes TTL but it makes sense to update `cache_expiration_time` here too.

Probably should be a separate commit from the test fix, though.

src/SimplePie.php
```diff
@@ -1924,6 +1924,7 @@ protected function fetch_data(&$cache)
                             $this->status_code = 0;
 
                             if ($this->force_cache_fallback) {
+                                $this->data['cache_expiration_time'] = $this->cache_duration + time();
```
jtojnar, line 1927: Also probably want to do the same for 304 response.

jtojnar (commented):
Thanks for this. I think it would probably be cleanest to include the -1 hack in the original PR since it originally fixed the bug, and leave this one for the (IIUC) separate bug of `cache_expiration_time` not being updated on `force_cache_fallback`304.

src/SimplePie.php
```diff
@@ -1900,7 +1900,7 @@ protected function fetch_data(&$cache)
                     $this->data = [];
                 }
                 // Check if the cache has been updated
-                elseif (isset($this->data['cache_expiration_time']) && $this->data['cache_expiration_time'] > time()) {
+                elseif (empty($this->data['cache_expiration_time']) || $this->data['cache_expiration_time'] < time()) {
```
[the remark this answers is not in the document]
Alkarex: Fair preference for `isset` here but not `array_key_exists()` as there would be a `null` vs. `int` inequality right after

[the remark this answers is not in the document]
Alkarex: Regarding `&&`, rather not. If for any reason `cache_expiration_time` is not set, I would want my cache to expire as soon as possible

src/SimplePie.php
Alkarex: The tests are still related to cache, and were preventing this PR from passing the tests, but can be put in another PR

src/SimplePie.php
jtojnar: Oh, I was wondering why does not the in-cache TTL remove the entry when it expires, which would break the 304 and it does, as you found out in https://github.com/FreshRSS/simplepie/blob/5e92ce767597aa9f5cbb00d88c7591b2e604dc5f/src/Cache/BaseDataCache.php#L58-L62

Alkarex: I can happily send upstream more of the fixes / improvements, but best after some of the existing pending PRs are processed

src/SimplePie.php
Art4: @jtojnar `$this->data['cache_expiration_time']` was introduced to allow mimic the functionality of `mtime()` and `touch()`, because that allowed us to use the cached data even if it was expired. The problem

jtojnar: Hmm, maybe we should just pass TTL as `null` (unlimited) to Psr16 caches, and have another layer that would emulate the old `mtime()`/`touch()` API, if we want to have access to expired cache entries. I do not see how can we fix the cache fallback otherwise.

In the long term, we can probably come up with a nicer API though. For example, we could have a class `CacheWrapper` that would handle expiration and invalidation transparently – it could store the cache entry as follows

```php
/** @type CacheEntry array{mtime: int, extraKeys: array<string, mixed>, value: mixed} */
```

and be used something like this:

```php
$cacheWrapper = new CacheWrapper(new Psr16($this->cache), $this->cache_duration);

$this->data = $cacheWrapper->get(
    $key,
    // Extra validation keys, since the allowed size of `$key` is limited.
    // If value of any of the keys does not match the one in the `extraKeys` array,
    // the value will be considered stale and one of the following actions will be performed.
    [
        // Different version of SimplePie might have unrecognizable structure of the cache data
        // so pass `null` as `$staleData` to the callback.
        ['key' => 'build', 'expected' => Misc::get_build(), 'action' => CacheWrapper::INVALID_ACTION_DISCARD],
        // A collision. Cause the value returned by the callback to be returned by `CacheWrapper::get()` but the cache would not be updated.
        // I guess this could also be handled transparently by having the `CacheWrapper` store it in `CacheEntry`.
        ['key' => 'url', 'expected' => $this->feed_url, 'action' => CacheWrapper::INVALID_ACTION_COLLISION],
        // …
    ],
    // This callback will be executed when the cache lacks the key, or it is invalid.
    // `$staleData` will be null if the key did not exist.
    // The cache will be updated with the returned value and the mtime will be updated.
    // Except maybe when exception occurs, then we might want to unset the cache and propagate the exception.
    function(?array $staleData) use (&$obtainedFromCache): array {
        $headers = [];
        if ($staleData !== null && isset($staleData['headers']['etag'])) {
            $headers['etag'] = $staleData['headers']['etag'];
        }

        $file = fetch($url, $headers);

        if ($this->status_code === 304) {
            $obtainedFromCache = true;
            return $staleData;
        }

        // …

        return  [
            'url' => $this->feed_url,
            'feed_url' => $file->get_final_requested_uri(),
            'build' => Misc::get_build(),
            'cache_expiration_time' => $this->cache_duration + time(),
        ];
    }
);
```

Alkarex: I would suggest to address this potential refactoring in another PR though :-)
The current PR fixes some relatively obvious bugs, so merging it would already be beneficial

Alkarex: P.S. Regarding the tests, the change is due to the fix of the inequality higher up, so they are directly related

Alkarex:
@Art4 As far as this PR goes (and excluding future fixes / refactoring), isn't it also good from your point of you?

src/SimplePie.php
```diff
@@ -1900,7 +1900,7 @@ protected function fetch_data(&$cache)
                     $this->data = [];
                 }
                 // Check if the cache has been updated
-                elseif (isset($this->data['cache_expiration_time']) && $this->data['cache_expiration_time'] > time()) {
+                elseif (!isset($this->data['cache_expiration_time']) || $this->data['cache_expiration_time'] < time()) {
```
jtojnar, line 1903: Reading the explanations from both of you I think this PR is reasonable.

The last remaining thing not clear to me is how it can this patch fix anything without https://github.com/FreshRSS/simplepie/blob/5e92ce767597aa9f5cbb00d88c7591b2e604dc5f/src/Cache/BaseDataCache.php#L58-L62

If we use the same `$this->cache_duration` as TTL passed to `$cache->set_data()` and to `$this->data['cache_expiration_time']`, I would expect that if the cache expires, the following would return the default empty array, and the above line would never be executed, as it is within `if (!empty($this->data))`:

https://github.com/simplepie/simplepie/blob/6b97dec75ed45c11a7411616643f855fe13c2add/src/SimplePie.php#L1874

Alkarex: > how it can this patch fix anything

This patch will fix some cases when the internal TTL has not expired yet. I think this is needed in any case.

But I completely agree that to have a working 304 cache, more fixes of the logic are needed. This PR focusses on the straightforward bugs in the current logic, which are needed no matter what.

jtojnar: But how could the internal TTL not have expired when `$this->data['cache_expiration_time']` did – are not they set to the same cache duration?

Art4 (approved):

src/SimplePie.php
Alkarex: While the default cache implementation ("legacy file cache") is still broken, I believe passing another implementation might work (but I have not investigated in details)

Alkarex:
@jtojnar With all the commits in this PR, and associated approvals and comments, I believe a *squash and merge* would be good, making sure to put in the commit message for instance the text describing this PR all the way up (which I have just edited).
But as you prefer, I do not really mind 🙂

src/SimplePie.php
jtojnar: Ah, alright. We can set the TTL to `null` in all `set_data` calls in a follow up PR to actually fix it then.

jtojnar (approved):
Yes, squash merge makes sense here.

I propose something like the following commit message:

> Set cache_expiration_time when storing data to cache
>
> When we switched to PSR-16 inspired `DataCache` in bb38d740c12a953dce4e2722a8d26d7a3ed2be3f, we introduced a regression due to impedance mismatch between PSR-16 and HTTP Caching mechanisms:
> - HTTP caching uses two step invalidation – if a cache entry is accessed after its TTL passes, a request is made with a stored HTTP caching header (e.g. ETag) and the cache is considered invalid only if the server returns something else than 304 Not Modified response.
> - On the other hand, PSR-16 cache does not allow access to entries past its TTL, forcing us to fetch the content every time, even when the remote server could tell us it is still fine.
>
> To work around this, bb38d740c12a953dce4e2722a8d26d7a3ed2be3f decided to store `cache_expiration_time` field in the cache entry so that we could bypass the PSR-16’s strict TTL and handle expiration ourselves. But there were few places where we forgot to set it. This patch fixes that.
>
> Unfortunately, because we also pass the same value as `cache_expiration_time` to `DataCache::set_data` as TTL, the early eviction still happens. We will also need to set the TTL to `null` (unlimited) so that the `DataCache` does not evict the entries behind our back.
>
> Partly fixes https://github.com/simplepie/simplepie/issues/830 ([BUG] Cache constantly updating)

This the context according to my understanding, hope I did not miss something.

PR #875, opened 2024-08-27
[1.8] Backport PHP 8.4 compat fixes

jtojnar:

PR #886, opened 2024-09-25
Release 1.8.1

jtojnar:
To be backported:
- [x] https://github.com/simplepie/simplepie/pull/887
- [x] #876 
- [ ] #883
- [x] https://github.com/simplepie/simplepie/pull/894

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased](https://github.com/simplepie/simplepie/compare/1.8.0...master)
+## [Unreleased](https://github.com/simplepie/simplepie/compare/1.8.1...one-dot-eight)
+
+Nothing yet.
+
+## [1.8.1](https://github.com/simplepie/simplepie/compare/1.8.0...1.8.1) - 2024-10-01
+
+### Changed
+
+- Disable Composer lockfile by @jrfnl in [#887](https://github.com/simplepie/simplepie/pull/887), backported in [#895](https://github.com/simplepie/simplepie/pull/895)
+
+### Fixed
+
+- Fixes for PHP 8.4 deprecations by @Girgias, @jrfnl and @jtojnar, backported in [#875](https://github.com/simplepie/simplepie/pull/875)
+- Fix locator with website missing `Content-Type` header by @jtojnar in [#891](https://github.com/simplepie/simplepie/pull/891)
+- Fix `encode` argument of `SimplePie::strip_htmltags()` @jtojnar in [#894](https://github.com/simplepie/simplepie/pull/894), backported in [#898](https://github.com/simplepie/simplepie/pull/898)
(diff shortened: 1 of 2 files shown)


Alkarex:
The cache is broken at the moment. Requires https://github.com/simplepie/simplepie/pull/883

CHANGELOG.md
```diff
@@ -5,7 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased](https://github.com/simplepie/simplepie/compare/1.8.0...master)
```
Art4, line 8: We should keep the `Unreleased` section for future notes and just update the tag from 1.8.0 to 1.8.1

See also https://keepachangelog.com/en/1.1.0/#effort

Art4 (changes requested):

CHANGELOG.md
jtojnar: I find it weird to have stubs like this in a final release. I could add it in a separate commit and then tag the first commit.

Or maybe it does not matter.

Art4: For the newest release you are right. But with more time this section becomse more important. One user with the e.g. 1.7.0 release has now a link to all new changes that one will get with an update.

See https://github.com/simplepie/simplepie/blob/1.7.0/CHANGELOG.md

jtojnar: I do not see much of a point – you typically update to a specific version, so you would check the changelog for the version you are updating to and if you want to see commits, you would look at the diffs version by version. If you really want all commits at once, you can modify the URL or use `git log` but that should be rare. Either way, seeing all commits until `HEAD` will give you extra commits past the last release, which are not relevant unless you are updating to `HEAD`.

Also, once next version rolls in, the “Unreleased” link label is no longer accurate.

But it is not that important, it is only mildly confusing so until https://github.com/olivierlacan/keep-a-changelog/discussions/517 is resolved, I am fine with following what [keep-a-changelog project does](https://github.com/olivierlacan/keep-a-changelog/blob/v1.1.1/CHANGELOG.md) and not bother removing it.

jrfnl:
The release should be blocked until the fix to `Sanitize` from PR #880 has been backported.

Also see: https://github.com/simplepie/simplepie/pull/875#issuecomment-2375611125

CHANGELOG.md
jrfnl: If it helps, in other projects, I often see the "Unreleased" header (and link) to remain on releases, but accompanied by one extra sentence underneath the header: _Nothing yet._

Art4:
I would like to get Release 1.8.1 ready for release.

@jtojnar I've created backports for #887 and #876. 

#889 cannot be backported because PHPStan was introduced in #811 on the `master` branch.

Before backporting #883, the problem with `$ttl = null` must be solved first.

jtojnar:
Thanks. I have merged them.

Also opened [one more](https://github.com/simplepie/simplepie/pull/898), put back the unreleased section, added the suggested paragraph, and updated the release notes.

Do we want to block the release on the cache fix?

Art4:
> Do we want to block the release on the cache fix?

Yes, I would should fix this first.

Art4 (approved):

jtojnar:
Thanks. Pushed a tag.

PR #887, opened 2024-09-26
Composer: prevent a lock file from being created

jrfnl:
Composer 1.10.0 introduced a `lock` config option, which, when set to `false` will prevent a `composer.lock` file from being created and will ignore it when one exists.

This is a useful option for PHP cross-version compatible packages such as this, where the `lock` file has no meaning.

It also makes life more straight-forward for contributors as they don't have to remember that for this repo they should use `composer update` instead of `composer install`. Both will now work the same.

Refs:
https://getcomposer.org/doc/06-config.md#lock

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,6 +61,7 @@
     },
     "config": {
         "bin-dir": "bin",
+        "lock": false,
         "sort-packages": true
     },
     "scripts": {


jrfnl:
Open question: should this be backported to the 1.8 branch ?

jtojnar (approved):
Thanks, sounds reasonable to me. Should be fine to backport too – I assume older versions of Composer will just ignore the unknown key.

jrfnl:
> Thanks, sounds reasonable to me. Should be fine to backport too – I assume older versions of Composer will just ignore the unknown key.

Composer 1.10 has been out for a long time and is barely used anymore. Packagist is even [deprecating support for Composer 1.x](https://blog.packagist.com/shutting-down-packagist-org-support-for-composer-1-x/) and support will be removed completely by August 2025, so I don't think support for Composer < 1.10 should be a concern.

Art4:
LGTM. We should also remove the `composer.lock` entry from the `.gitignore` file.

jrfnl:
> LGTM. We should also remove the `composer.lock` entry from the `.gitignore` file.

I left it in for a reason - existing contributors may still have the file in their local clone and even though Composer will now ignore it, it would probably still not be a good idea for (an outdated version of) `composer.lock` to accidentally get committed.

PR #892, opened 2024-09-30
PHPStan: Add extension for typing Registry::call()

jtojnar:
`Registry::call()` is widely used in SimplePie to allow extensibility. Unfortunately, the pattern hides the actual method from PHPStan so it cannot determine the return value type, and there is no annotation we can add to make it work. , This results in an inadequate analysis.

Thankfully, most cases allow us to determine the method statically so we can create a PHPStan extension doing just that.

The extension might be useful for SimplePie consumers so let’s expose it in `composer.json`.

We may want to add more type checking extensions in the future.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,6 +56,7 @@
     },
     "autoload-dev": {
         "psr-4": {
+            "SimplePieUtils\\": "utils/",
             "SimplePie\\Tests\\Fixtures\\": "tests/Fixtures",
             "SimplePie\\Tests\\Unit\\": "tests/Unit"
         }
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -65,3 +65,6 @@ parameters:
             message: '(^Access to an undefined property XMLReader::\$\w+\.$)'
             # Only occurs on PHP ≥ 8.2
             reportUnmatched: false
+
+includes:
+    - utils/PHPStan/extension.neon
--- a/src/Misc.php
+++ b/src/Misc.php
@@ -1858,7 +1858,7 @@ class Misc
 
     /**
      * @param array<string, array<string, string>> $attribs
-     * @return int
+     * @return int-mask-of<SimplePie::CONSTRUCT_*>
      */
     public static function atom_10_construct_type(array $attribs)
     {
@@ -1882,7 +1882,7 @@ class Misc
 
     /**
      * @param array<string, array<string, string>> $attribs
-     * @return int
+     * @return int-mask-of<SimplePie::CONSTRUCT_*>
      */
     public static function atom_10_content_construct_type(array $attribs)
     {
--- a/utils/PHPStan/README.md
+++ b/utils/PHPStan/README.md
@@ -1,0 +1,7 @@
+# SimplePie PHPStan extension
+
+**This extension is considered unstable, use at your own risk.**
+
+It provides:
+
+- Correct return type for `Registry::call()`
--- a/utils/PHPStan/RegistryCallMethodReturnTypeExtension.php
+++ b/utils/PHPStan/RegistryCallMethodReturnTypeExtension.php
@@ -1,0 +1,101 @@
+<?php
+
+// SPDX-FileCopyrightText: 2024 Jan Tojnar
+// SPDX-License-Identifier: BSD-3-Clause
+
+declare(strict_types=1);
+
+namespace SimplePieUtils\PHPStan;
+
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Type\DynamicMethodReturnTypeExtension;
+use PHPStan\Type\Type;
+use PHPStan\Reflection\ReflectionProvider;
+use PhpParser\Node\Expr\MethodCall;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\Constant\ConstantArrayType;
+use PHPStan\Type\ArrayType;
+use PHPStan\Type\MixedType;
+
+/**
+ * Fixes return type for `Registry::call()` to match the called method.
+ */
+class RegistryCallMethodReturnTypeExtension implements DynamicMethodReturnTypeExtension
+{
+    /** @var ReflectionProvider */
+    private $reflectionProvider;
+
+    public function __construct(ReflectionProvider $reflectionProvider)
+    {
+        $this->reflectionProvider = $reflectionProvider;
+    }
+
+    public function getClass(): string
+    {
+        return 'SimplePie\Registry';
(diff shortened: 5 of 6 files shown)


Art4:
I would reccomend to move this extension class inside the tests folder, maybe `tests/Fixtures/PHPStan` and namespace `SimplePie\Tests\Fixtures\PHPStan`.

jtojnar:
I do not think it is really a fixture. I am also considering adding more maintenance scripts (e.g. [Rector](https://getcomposer.org/doc/articles/vendor-binaries.md#can-vendor-binaries-be-installed-somewhere-other-than-vendor-bin-) rules for automatic migration to SimplePie 2).

Art4:
> I do not think it is really a fixture. I am also considering adding more maintenance scripts (e.g. [Rector](https://getcomposer.org/doc/articles/vendor-binaries.md#can-vendor-binaries-be-installed-somewhere-other-than-vendor-bin-) rules for automatic migration to SimplePie 2).

While reading the rector book I noticed that rector recommends `utils/rector/` for custom rules. So I recommend moving the PHPStan files into `utils/phpstan/`.

> ## [File Structure](https://getrector.com/documentation/custom-rule#content-file-structure)
> 
> This is how the file structure for custom rule in your own project will look like:
> ```
> 
> /src
>     SomeCode.php
> 
> /utils
>     /rector
>         /src
>             /Rector
>                 MyFirstRector.php
> 
>         /tests
>             /Rector
>                 /MyFirstRector
>                     /Fixture
>                         test_fixture.php.inc
>                     /config
>                         config.php
>                     MyFirstRectorTest.php
> 
> rector.php
> composer.json
> ```

Art4 (changes requested):
LGTM, but we should move the `src/PHPStan` files into `utils/phpstan/`. This way we won't need to maintain the `build/compile.php` and are open for future rector rules.

Art4:
The [suggested namespaces](https://getrector.com/documentation/custom-rule#content-update-composerjson) are:

```
{
    "autoload": {
        "psr-4": {
            "App\\": "src"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "Utils\\Rector\\": "utils/rector/src",
            "Utils\\Rector\\Tests\\": "utils/rector/tests"
        }
    }
}
```

I've looked into SimplePie folders and namespaces (ignoring the `library` folder):

| Folder | Namespace |
|--------|--------|
| `src/` | `SimplePie\` |
| `tests/Integration/` | `SimplePie\Tests\Integration\` |
| `tests/Fixtures/` | `SimplePie\Tests\Fixtures\` |
| `tests/Unit/` | `SimplePie\Tests\Unit\` |

Following this pattern the `utils` would have this namespace:

| Folder | Namespace |
|--------|--------|
| `utils/` | `SimplePie\Utils\` |
| `utils/PHPStan/` | `SimplePie\Utils\PHPStan\` |
| `utils/PHPStan/Extension/` | `SimplePie\Utils\PHPStan\Extension\` |

But I don't like that we restrict us from using the potential `SimplePie\Utils\` in the `src` folder.

What do you think about moving the tests into a separate `SimplePieTests` namespace and the utils would go into the `SimplePieUtils` namespace?

| Folder | Namespace |
|--------|--------|
| `src/` | `SimplePie\` |
| **Tests** |
| `tests/` | `SimplePieTests\` |
| `tests/Integration/` | `SimplePieTests\Integration\` |
| `tests/Fixtures/` | `SimplePieTests\Fixtures\` |
| `tests/Unit/` | `SimplePieTests\Unit\` |
| `utils/` | `SimplePieUtils\` |
| **PHPStan extension** |
| `utils/PHPStan` | `SimplePieUtils\PHPStan\` |
| `utils/PHPStan/Extension` | `SimplePieUtils\PHPStan\Extension\` |
| **Potential future Rector rules** |
| `utils/Rector` | `SimplePieUtils\Rector\` |
| `utils/Rector/Rules` | `SimplePieUtils\Rector\Rules\` |
| `utils/Rector/Tests` | `SimplePieUtils\Rector\Tests\` |

The PSR-4 rules could be condensed into:

```json
{
    "autoload": {
        "psr-4": {
            "SimplePie\\": "src"
        },
        "psr-0": {
            "SimplePie": "library"
        }
    },
    "autoload-dev": {
        "psr-4": {
            "SimplePieTests\\": "tests",
            "SimplePieUtils\\": "utils"
        }
    },
}
```

jtojnar:
Since tests are internal only, the `SimplePie` prefix is probably redundant. We could remove it.

I guess they apply the same logic to the internal Rector rules, putting them into `Utils`.

Looking at some third-party rule sets, `RectorSimplePie` would be in line:

https://github.com/efabrica-team/rector-nette/blob/40bb22059b380beadae7137b0c20976832916eb2/composer.json#L36
https://github.com/driftingly/rector-laravel/blob/8eaa90a1c62a5a133228dd12d53935eadd0300b8/composer.json#L26

While first-party ones use `Rector\\` prefix:

https://github.com/rectorphp/rector-doctrine/blob/0231850ae962c82ff48adc26941475d6fa1b03dc/composer.json#L28
https://github.com/rectorphp/rector-symfony/blob/158a72ea131045f5be7c4ec81beab1938c737277/composer.json#L34


PHPStan appears to follow the same convention for first-party packages:

https://github.com/phpstan/phpstan-phpunit/blob/22c6949f5c0d4d3d343fbd42d2bd75472089d135/composer.json#L35
https://github.com/rectorphp/rector-doctrine/blob/0231850ae962c82ff48adc26941475d6fa1b03dc/composer.json#L28

but the third party ones are bit more wild:

https://github.com/Jan0707/phpstan-prophecy/blob/9c5f67d46d88fb10dd5e674331a35550d51b626c/composer.json#L31
https://github.com/larastan/larastan/blob/b9159cf7f75a668f6341895cbe8d4dcbebd3a775/composer.json#L53
https://github.com/szepeviktor/phpstan-wordpress/blob/4edada80613f627fd320b56b762706a7d4a53b90/composer.json#L34
https://github.com/mglaman/phpstan-drupal/blob/42cca54ee8bccec83a674ac45d1d17586777187e/composer.json#L41

So it really depends on if we want to to expose the extension publicly or not. If yes, `SimplePieUtils` as you propose sounds best to me.

Though it will probably need to be part of non-`dev` autoload, unless Composer includes `autoload-dev` of packages in `require-dev`. And at that point, we might as well put it in `src/`. Yes, it will require an extra ignore line in the `build.php` but I am not hiding [my derision for that](https://github.com/simplepie/simplepie/pull/893).

jtojnar:
- Removed mention of the extension from README.
- Removed phpstan installer support.
- Moved to `utils/`
- Switched namespace to `SimplePieUtils\PhpStan`.
- Added readme explaining the extension is unstable.

Art4 (approved):

PR #900, opened 2024-11-04
Update date parsing to use DateTimeImmutable

pbearne:
Replaced `strtotime` with `DateTimeImmutable` for more accurate and reliable date conversion. This change affects both `Parser.php` and `Parse/Date.php` and enhances the handling of date strings.

### DIFF
--- a/src/Parse/Date.php
+++ b/src/Parse/Date.php
@@ -942,7 +942,7 @@ PCRE;
     }
 
     /**
-     * Parse dates using strtotime()
+     * Parse dates using DateTimeImmutable()
      *
      * @access protected
      * @param string $date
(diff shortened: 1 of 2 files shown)


jtojnar:
Thanks for the contribution.

1. Could you describe which formats does the this support that `strtotime` does not? Could you include tests for them?
2. Are there any formats that will no longer be supported? Perhaps we should add this as a separate method with [`date_` prefix](https://github.com/simplepie/simplepie/blob/52c764f1db8c21fec7d598210e24d27f0aae83bf/src/Parse/Date.php#L609) so that it is used as a fallback when `strtotime` fails?

PR #901, opened 2024-11-22
GH Actions: PHP 8.4 has been released

jrfnl:
* Builds against PHP 8.4 are no longer allowed to fail.

Ref: https://www.php.net/releases/8.4/en.php


jtojnar:
Thanks.

PR #909, opened 2025-01-14
Rename HttpException to ClientException

Art4:
This PR renames the `SimplePie\Exception\HttpException` to `SimplePie\HTTP\ClientException` to avoid confusion that this exception is thrown on an 4xx or 5xx response from the server. Because the `ClientException` is always catched internally I marked the class as `@internal`.

This PR also adds tests to show that the responses (`File` or `Psr7Response` class) always contains the status code, even if it is 500.

### DIFF
--- a/tests/Integration/HTTP/ClientsTest.php
+++ b/tests/Integration/HTTP/ClientsTest.php
@@ -11,8 +11,8 @@ use PHPUnit\Framework\TestCase;
 use Psr\Http\Client\ClientInterface;
 use Psr\Http\Message\RequestFactoryInterface;
 use Psr\Http\Message\UriFactoryInterface;
-use SimplePie\Exception\HttpException;
 use SimplePie\HTTP\Client;
+use SimplePie\HTTP\ClientException;
 use SimplePie\HTTP\FileClient;
 use SimplePie\HTTP\Psr18Client;
 use SimplePie\HTTP\Response;
@@ -52,16 +52,16 @@ class ClientsTest extends TestCase
         $this->assertStringStartsWith('<rss version="2.0">', $response->get_body_content());
     }
 
-    public function testFileClientThrowsHttpException(): void
+    public function testFileClientThrowsClientException(): void
     {
-        $this->runTestWithClientThrowsHttpException(
+        $this->runTestWithClientThrowsClientException(
             new FileClient(new Registry())
         );
     }
 
-    public function testPsr18ClientThrowsHttpException(): void
+    public function testPsr18ClientThrowsClientException(): void
     {
-        $this->runTestWithClientThrowsHttpException(
+        $this->runTestWithClientThrowsClientException(
             new Psr18Client(
                 $this->createMock(ClientInterface::class),
                 $this->createMock(RequestFactoryInterface::class),
@@ -70,11 +70,11 @@ class ClientsTest extends TestCase
         );
     }
 
-    private function runTestWithClientThrowsHttpException(Client $client): void
+    private function runTestWithClientThrowsClientException(Client $client): void
     {
         $filepath = dirname(__FILE__, 3) . '/data/this-file-does-not-exist';
 
-        $this->expectException(HttpException::class);
+        $this->expectException(ClientException::class);
         $this->expectExceptionCode(0);
 
         $this->expectExceptionMessage(sprintf('file "%s" is not readable', $filepath));
--- a/tests/Unit/HTTP/FileClientTest.php
+++ b/tests/Unit/HTTP/FileClientTest.php
@@ -75,4 +75,32 @@ final class FileClientTest extends TestCase
             ['Accept' => 'application/atom+xml']
         );
     }
+
+    public function testFileClientReturnsResponseWithStatusCode500(): void
+    {
+        $response = $this->createStub(File::class);
+        $response->method('get_status_code')->willReturn(500);
+
+        $registry = $this->createStub(Registry::class);
+        $registry->method('create')->willReturn($response);
+
+        $client = new FileClient($registry, [
+            'timeout' => 30,
+            'useragent' => 'dummy-useragent',
+            'redirects' => 10,
+            'force_fsockopen' => true,
+            'curl_options' => [
+                \CURLOPT_FAILONERROR => true,
+            ],
+        ]);
+
+        $response = $client->request(
+            FileClient::METHOD_GET,
+            'http://example.com/500-error',
+            ['Accept' => 'application/atom+xml']
+        );
+
+        // Make sure no ClientException is thrown on status code 500
+        $this->assertSame(500, $response->get_status_code());
+    }
 }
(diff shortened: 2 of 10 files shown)


tests/Unit/HTTP/FileClientTest.php
```diff
@@ -75,4 +75,32 @@ public function testFileClientProvidesCorrectOptionsAndHeaders(): void
...
+
+        $client = new FileClient($registry, [
+            'timeout' => 30,
+            'useragent' => 'dummy-useragent',
+            'redirects' => 10,
+            'force_fsockopen' => true,
+            'curl_options' => [
+                \CURLOPT_FAILONERROR => true,
+            ],
+        ]);
```
jtojnar, line 95: What is the point of setting the options instead of defaults? Also the last two are mutually exclusive.

```suggestion
        $client = new FileClient($registry);
```

tests/Unit/HTTP/FileClientTest.php
```diff
@@ -75,4 +75,32 @@ public function testFileClientProvidesCorrectOptionsAndHeaders(): void
             ['Accept' => 'application/atom+xml']
         );
     }
+
+    public function testFileClientReturnsResponseWithStatusCode429(): void
+    {
+        $response = $this->createStub(File::class);
+        $response->method('get_status_code')->willReturn(429);
```
jtojnar, line 82: This test is misleading.

Thanks to [`CURLOPT_FAILONERROR`](https://github.com/simplepie/simplepie/blob/452f47906f59a87ddda42d0dcf74e58403372bb3/src/File.php#L130), [`curl_errno`](https://github.com/simplepie/simplepie/blob/452f47906f59a87ddda42d0dcf74e58403372bb3/src/File.php#L145-L148) will return [`CURLE_HTTP_RETURNED_ERROR`](https://www.php.net/manual/en/curl.constants.php#constant.curle-http-returned-error), `File::$success` will be `false`, and `FileClient` will [still throw an error](https://github.com/simplepie/simplepie/blob/452f47906f59a87ddda42d0dcf74e58403372bb3/src/HTTP/FileClient.php#L72-L74):

```php
<?php

declare(strict_types=1);

use SimplePie\Registry;
use SimplePie\HTTP\FileClient;
use SimplePie\HTTP\Psr18Client;

require __DIR__ . '/vendor/autoload.php';

$client = new Psr18Client(
    new \GuzzleHttp\Client(),
    new \GuzzleHttp\Psr7\HttpFactory(),
    new \GuzzleHttp\Psr7\HttpFactory()
);
var_dump($client->request('GET', 'https://httpbin.org/status/403')->get_status_code());

$registry = new Registry();
$client = new FileClient($registry);
var_dump($client->request('GET', 'https://httpbin.org/status/403')->get_status_code());
```

```shell
$ php test.php
/home/jtojnar/Projects/simplepie/test.php:17:
int(403)

Fatal error: Uncaught SimplePie\HTTP\ClientException: cURL error 22: The requested URL returned error: 403 in /home/jtojnar/Projects/simplepie/src/HTTP/FileClient.php on line 72

SimplePie\HTTP\ClientException: cURL error 22: The requested URL returned error: 403 in /home/jtojnar/Projects/simplepie/src/HTTP/FileClient.php on line 72

Call Stack:
    0.0003     471280   1. {main}() /home/jtojnar/Projects/simplepie/test.php:0
    0.6658    2902416   2. SimplePie\HTTP\FileClient->request($method = 'GET', $url = 'https://httpbin.org/status/403', $headers = ???) /home/jtojnar/Projects/simplepie/test.php:21
```

Maybe something like this for now:

```suggestion
        $this->markTestIncomplete('This test describes expected behavior but is currently broken.');

        $response = $this->createStub(File::class);
        $response->method('get_status_code')->willReturn(429);
        // `File` enables `CURLOPT_FAILONERROR`, so it will set these on HTTP error.
        $response->error = 'cURL error 22: The requested URL returned error: 403';
        $response->success = false;
```

jtojnar (commented):
Thanks, and sorry for the delay in the review.

The renaming is reasonable but the file test is misleading as `FileClient` behaves differently from `Psr18Client`.

tests/Unit/HTTP/FileClientTest.php
pe1uca: I'm not sure marking the test as incomplete is a good idea, having in mind is directly the intention of this PR, avoid throwing `ClientException`.

Art4: I would keep this test to make it clear that `FileClient` behaves similar to PSR-18 and doesn't throw an exception on any response.

The property values are tested in the integration test. They are not necessary in this case and I would keep a stub as minimal as possible.

tests/Integration/HTTP/ClientsTest.php
```diff
@@ -70,15 +72,69 @@ public function testPsr18ClientThrowsHttpException(): void
...
+        );
+
+        $response = $client->request(Client::METHOD_GET, $url);
+
+        $server->stop();
+
+        $this->assertSame(500, $response->get_status_code());
+    }
+
+    public function testFileClientThrowsClientExceptionIfServerIsUnreachable(): void
```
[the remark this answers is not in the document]
Art4: Nice to know, thank you! I'll change the host to `example.invalid`.

tests/Unit/HTTP/FileClientTest.php
jtojnar: Well, that really depends on what we want to cover with this test.

The code in question

https://github.com/simplepie/simplepie/blob/a6d49d359ce0f5373ebaccc26430102d0290cb36/src/HTTP/FileClient.php#L71

[condition coverage](https://en.wikipedia.org/wiki/Code_coverage#Basic_coverage_criteria) requires three states:

1. `$file->success` + any `$file->get_status_code()`
2. `!$file->success` + `$file->get_status_code() !== 0`
3. `!$file->success` + `$file->get_status_code() === 0`

Currently, `testFileClientReturnsResponseWithStatusCode429` seems to want to cover (2), but actually covers (1).

If we want to have an unit test that covers (3), that would be `testFileClientThrowsExceptionOnServerConnectionError`.

jtojnar (approved):
Looks perfect now. Suggest squash-merging with the following message:

```md
Rename HttpException to ClientException and stop throwing on HTTP error in FileClient

This PR renames the `SimplePie\Exception\HttpException` to `SimplePie\HTTP\ClientException` to avoid confusion (https://github.com/simplepie/simplepie/pull/905) that this exception is thrown on an 4xx or 5xx response from the server. Because the `ClientException` is always caught internally, I marked the class as `@internal`.

Additionally, change the `FileClient` to only throw the exception on non-HTTP errors to match PSR-18 and our `Psr18Client`. This is not a BC break since this API is not part of any release yet.

Also add unit tests to show that the responses (`File` or `Psr7Response` class) always contain the status code, even if it is 429. Plus integration tests using a HTTP server.
```

PR #910, opened 2025-01-27
add development related config files to `.gitattributes`

ziegenberg:
Those files should not end up in release packages.


jtojnar:
Thanks.

PR #911, opened 2025-02-17
Update Date.php

Juergen-W:
Method call for implode used the legacy signaturr which is no longer supported with PHP 8.0+x.

### DIFF
--- a/src/Parse/Date.php
+++ b/src/Parse/Date.php
@@ -556,8 +556,8 @@ class Date
      */
     public function __construct()
     {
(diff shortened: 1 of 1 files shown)


jtojnar (changes requested):
According to https://www.php.net/manual/en/function.implode.php, your proposal is actually the deprecated syntax.

Juergen-W:
Sorry, you are right - i was confused because i may get an outdated version from https://simplepie.org/downloads/?download.

PR #912, opened 2025-03-05
FileClient: Merge HTTP headers from cURL options

Alkarex:
The `$curl_options` parameter could override the `CURLOPT_HTTPHEADER` option, so make sure to merge the HTTP headers properly.
Before this patch, providing an additional HTTP header via `$curl_options[CURLOPT_HTTPHEADER]` would remove all essential headers such as Accept, if-modified-since, etc.

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -115,6 +115,12 @@ class File implements Response
                 foreach ($headers as $key => $value) {
                     $headers2[] = "$key: $value";
                 }
+                if (isset($curl_options[CURLOPT_HTTPHEADER])) {
+                    if (is_array($curl_options[CURLOPT_HTTPHEADER])) {
+                        $headers2 = array_merge($headers2, $curl_options[CURLOPT_HTTPHEADER]);
+                    }
+                    unset($curl_options[CURLOPT_HTTPHEADER]);
+                }
                 if (version_compare(\SimplePie\Misc::get_curl_version(), '7.10.5', '>=')) {
                     curl_setopt($fp, CURLOPT_ENCODING, '');
                 }


jtojnar (approved):
Thanks.

This makes sense to me: AFAICT, `set_curl_options` is the only way besides subclassing to pass extra headers with requests, if a bit indiscriminate, so we should support it. And we probably do not want to discard headers set by SimplePie, as it should, in theory, know what it is doing.

Generally, it might be desirable to provide a mechanism for overriding the headers set by SimplePie (the way that can be used for removing headers set by cURL – setting a header to an empty value – does not work, duplicate headers from `$headers` and `$curl_options` will be sent), but since the only SimplePie only ever sets one of four headers (`Accept`, `if-modified-since`, `if-none-match`, `X-FORWARDED-FOR`), implementing it would probably not be worth it. Especially when the `FileClient` is meant to away eventually.

<details>
<summary>Comparison of the behavior</summary>

## Test script

```php
<?php

require __DIR__ . '/vendor/autoload.php';

$file = new SimplePie\File(
    'https://httpbin.org/headers',
    headers: [
        'X-Both-Args' => 'Headers',
        'X-Headers-Arg' => '1',
        'X-Headers-Arg-Hopefully-Shadowed' => 'Headers',
    ],
    curl_options: [
        \CURLOPT_HTTPHEADER => [
            'X-Both-Args: Options',
            'X-Curl-Options-Arg: 1',
            'X-Headers-Arg-Hopefully-Shadowed:',
            'X-Headers-Arg-Hopefully-Shadowed: Options',
        ]
    ]
);

echo $file->get_body_content();
```

## master

```json
{
  "headers": {
    "Accept": "*/*", 
    "Accept-Encoding": "deflate, gzip, br, zstd", 
    "Host": "httpbin.org", 
    "Referer": "https://httpbin.org/headers", 
    "X-Amzn-Trace-Id": "Root=1-67d54198-20e1c4c154bbe72c0db9357d", 
    "X-Both-Args": "Options", 
    "X-Curl-Options-Arg": "1", 
    "X-Headers-Arg-Hopefully-Shadowed": "Options"
  }
}
```

## This PR

```json
{
  "headers": {
    "Accept": "*/*", 
    "Accept-Encoding": "deflate, gzip, br, zstd", 
    "Host": "httpbin.org", 
    "Referer": "https://httpbin.org/headers", 
    "X-Amzn-Trace-Id": "Root=1-67d54136-2cf0ca7c20c8e2cd2ae221cc", 
    "X-Both-Args": "Headers,Options", 
    "X-Curl-Options-Arg": "1", 
    "X-Headers-Arg": "1", 
    "X-Headers-Arg-Hopefully-Shadowed": "Headers,Options"
  }
}
```

</details>

PR #914, opened 2025-04-18
Fix support for feeds with XML preample + DTD

Alkarex:
Simple feeds with XML preample + DTD are crashing (see example in test) due to regression from https://github.com/simplepie/simplepie/commit/162a3d3878b755c93820ea672d83b3f6ac866f8f

The culprit feature *added support for html entities in xml* seems to come from a misunderstanding of
https://www.rssboard.org/rss-encoding-examples

Example of feed: https://blog.plover.com/index.rss
Downstream Issue: https://github.com/FreshRSS/FreshRSS/issues/7514


Alkarex:
Although the regression is old in SimplePie, this was not spotted before because we used a patched version of SimplePie that had (by choice) not incorporated those changes until FreshRSS 1.25.0 released in December.
https://github.com/FreshRSS/FreshRSS/pull/4374

PR #915, opened 2025-06-23
Fix support for feeds with XML preample + DTD, entities

Alkarex:
Alternative of https://github.com/simplepie/simplepie/pull/914 with support for XML feeds using HTML entities

Simple feeds with XML preample + DTD are crashing (see example in test) due to regression from https://github.com/simplepie/simplepie/commit/162a3d3878b755c93820ea672d83b3f6ac866f8f
Its implementation is buggy, as it creates a second DOCTYPE declaration even when there is an existing one, which is completely invalid.

Example of feed: https://blog.plover.com/index.rss
Downstream Issue: https://github.com/FreshRSS/FreshRSS/issues/7514
Downstream PR: https://github.com/FreshRSS/simplepie/pull/35

Note that the culprit feature *added support for html entities in xml* seems to come from a misunderstanding of
https://www.rssboard.org/rss-encoding-examples
HTML entities are only allowed in XML when there is a DTD declaring them. SimplePie is even allowing Atom documents with undeclared HTML entities - which I am not sure is on purpose.

Downstream issue: https://github.com/FreshRSS/FreshRSS/issues/7687


src/Parser.php
```diff
@@ -111,11 +111,18 @@ public function parse(string &$data, string $encoding, string $url = '')
             $declaration = $this->registry->create(DeclarationParser::class, [substr($data, 5, $pos - 5)]);
             if ($declaration->parse()) {
                 $data = substr($data, $pos + 2);
-                $data = '<?xml version="' . $declaration->version . '" encoding="' . $encoding . '" standalone="' . (($declaration->standalone) ? 'yes' : 'no') . '"?>' ."\n". $this->declare_html_entities() . $data;
+                $data = preg_replace('/^\\s*<!DOCTYPE\\s[^>\\[\\]]*>\s*/', '', $data);  // Strip DOCTYPE except if containing an [internal subset]
+                $data = '<?xml version="' . $declaration->version . '" encoding="' . $encoding . '" standalone="' . (($declaration->standalone) ? 'yes' : 'no') . '"?>' . "\n" .
+                    (preg_match('/^\\s*<!DOCTYPE\\s/', $data) ? '' : $this->declare_html_entities()) .  // Declare HTML entities only if no remaining DOCTYPE
+                    $data;
```
jtojnar, line 117: Could this be extracted into a helper method:

```patch
diff --git a/src/Parser.php b/src/Parser.php
index 5c748477..feaf45e5 100644
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -111,18 +111,14 @@ class Parser implements RegistryAware
             $declaration = $this->registry->create(DeclarationParser::class, [substr($data, 5, $pos - 5)]);
             if ($declaration->parse()) {
                 $data = substr($data, $pos + 2);
-                $data = preg_replace('/^\\s*<!DOCTYPE\\s[^>\\[\\]]*>\s*/', '', $data);  // Strip DOCTYPE except if containing an [internal subset]
                 $data = '<?xml version="' . $declaration->version . '" encoding="' . $encoding . '" standalone="' . (($declaration->standalone) ? 'yes' : 'no') . '"?>' . "\n" .
-                    (preg_match('/^\\s*<!DOCTYPE\\s/', $data) ? '' : $this->declare_html_entities()) .  // Declare HTML entities only if no remaining DOCTYPE
-                    $data;
+                    $this->set_doctype($data);
             } else {
                 $this->error_string = 'SimplePie bug! Please report this!';
                 return false;
             }
         } else {
-            $data = preg_replace('/^\\s*<!DOCTYPE\\s[^>\\[\\]]*>\s*/', '', $data);   // Strip DOCTYPE except if containing an [internal subset]
-            $data = (preg_match('/^\\s*<!DOCTYPE\\s/', $data) ? '' : $this->declare_html_entities()) .  // Declare HTML entities only if no remaining DOCTYPE
-                $data;
+            $data = $this->set_doctype($data);
         }
 
         $return = true;
@@ -651,6 +647,15 @@ class Parser implements RegistryAware
         return true;
     }
 
+    private function set_doctype(string $data): string
+    {
+        // Strip DOCTYPE except if containing an [internal subset]
+        $data = preg_replace('/^\\s*<!DOCTYPE\\s[^>\\[\\]]*>\s*/', '', $data);
+        // Declare HTML entities only if no remaining DOCTYPE
+        $doctype = preg_match('/^\\s*<!DOCTYPE\\s/', $data) ? '' : $this->declare_html_entities();
+        return $doctype . $data;
+    }
+
     private function declare_html_entities(): string
     {
         // This is required because the RSS specification says that entity-encoded
```

jtojnar (commented):
Thanks, this looks reasonable, just a single nit.

tests/Unit/ItemTest.php
```diff
@@ -4312,6 +4312,93 @@ public static function getTitleDataProvider(): array
...
+		<item>
+			<title>Item Title</title>
+		</item>
+	</channel>
+</rss>
+XML
+                ,
+                'Item Title',
+            ],
+            'Test RSS 0.91 XML and DTD and external HTML entity' => [
```
jtojnar, line 4330: > HTML entity

The RSS 0.91 DTD actually copies the entity list from HTML 3.2 and appears to be a subset of  `declare_html_entities`.

tests/Unit/ItemTest.php
```diff
@@ -4312,6 +4312,93 @@ public static function getTitleDataProvider(): array
...
+<<<XML
+<?xml version="1.0"?>
+<!DOCTYPE rss PUBLIC "-//Netscape Communications//DTD RSS 0.91//EN" "http://www.rssboard.org/rss-0.91.dtd">
+<rss version="0.91">
+	<channel>
+		<item>
+			<title>caf&eacute;</title>
```
jtojnar, line 4337: Weirdly, the entities from `declare_html_entities` appear to be substituted, even though the `DOCTYPE` within uses `html` as root element. :woman_shrugging:

src/Parser.php
```diff
@@ -111,11 +111,18 @@ public function parse(string &$data, string $encoding, string $url = '')
...
             } else {
                 $this->error_string = 'SimplePie bug! Please report this!';
                 return false;
             }
+        } else {
+            $data = preg_replace('/^\\s*<!DOCTYPE\\s[^>\\[\\]]*>\s*/', '', $data);   // Strip DOCTYPE except if containing an [internal subset]
+            $data = (preg_match('/^\\s*<!DOCTYPE\\s/', $data) ? '' : $this->declare_html_entities()) .  // Declare HTML entities only if no remaining DOCTYPE
```
jtojnar, line 124: Instead of second matching, we could check the `$count` out-argument of `preg_replace`. But it probably does not matter in practice.

tests/Unit/ItemTest.php
```diff
@@ -4312,6 +4312,93 @@ public static function getTitleDataProvider(): array
...
+		</item>
+	</channel>
+</rss>
+XML
+                ,
+                'café',
+            ],
```
jtojnar, line 4401: In theory, the following should work as well but it is probably to niche to care about:

```suggestion
            ],
            'Test RSS 0.91 DTD and external HTML entity and internal entity' => [
<<<XML
<!DOCTYPE rss PUBLIC "-//Netscape Communications//DTD RSS 0.91//EN" "http://www.rssboard.org/rss-0.91.dtd" [
    <!ENTITY adios "Adi&oacute;s">
]>
<rss version="0.91">
    <channel>
        <item>
            <title>&adios; Adi&oacute;s</title>
        </item>
    </channel>
</rss>
XML
                ,
                'Adiós Adiós',
            ],
```

Alkarex: Indeed, but merging DTDs do not look easy as they may contain more complicated syntax

src/Parser.php
Alkarex: I am not sure how to do that, since we may have three different cases:
* No DTD at all
* A DTD that should be removed
* A DTD that should not be removed

tests/Unit/ItemTest.php
Alkarex: I changed to `rss`

PR #916, opened 2025-06-24
Fix feeds encoded in UTF-16LE

Alkarex:
Example of valid feed not working in SimplePie: https://haveibeenpwned.com/feed/breaches/
Regression due to https://github.com/simplepie/simplepie/pull/445
The final character `>` of a feed is encoded as `3E 00` in UTF-16LE, so calling `trim()` was removing the `\x00`, breaking the multibyte encoding and making the feed invalid.
Downstream issue https://github.com/FreshRSS/FreshRSS/issues/7690

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -238,7 +238,7 @@ class File implements Response
                                             $this->error = 'Unable to decode HTTP "gzip" stream';
                                             $this->success = false;
                                         } else {
-                                            $this->body = trim($decompressed);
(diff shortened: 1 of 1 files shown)


Alkarex:
P.S. That bug was surely also preventing UTF-32BE feeds from working, but not tested

jtojnar (approved):
LGTM, thanks. Tested with #819

Alkarex:
I just realised that the trimming was not uniformly applied to all cases.
Fixed and simplified in https://github.com/simplepie/simplepie/pull/917

PR #917, opened 2025-06-25
Trim body for all paths

Alkarex:
Slight refactor of https://github.com/simplepie/simplepie/pull/916 to cover all paths.
Missing paths included the fsock method without gzip (e.g. deflate or plain).

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -156,7 +156,7 @@ class File implements Response
                     $parser = new \SimplePie\HTTP\Parser($responseHeaders, true);
                     if ($parser->parse()) {
                         $this->set_headers($parser->headers);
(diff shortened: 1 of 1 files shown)


jtojnar:
Actually, thinking about it some more, the trimming does not really make sense with unknown encoding.

For example: With UTF-16BE, `\n` will be encoded as `000A`. If the file contains a final newline, as is [common on UNIX platforms](https://en.wikipedia.org/wiki/Newline#Interpretation), `trim` will mangle it into `00`, resulting in a half of a code point.

Less critically, with UTF-16LE, `€` will be encoded as `AC20`, but `20` is a space in ASCII, so `trim` will mangle file ending with € (though this is not a valid XML document, so it might no matter).

But even worse, arbitrary encoding can use arbitrary byte sequences to represent arbitrary code points. But again, not sure if we care about anything but ASCII-compatible encodings nowadays – we do not test any so they are probably broken anyway.

PR #918, opened 2025-06-26
tests/Integration: Add regression tests

jtojnar:
Adds regression test for #812.

Depends on https://github.com/donatj/mock-webserver/pull/75


tests/Integration/HTTP/BinaryResponse.php
Art4, line 1: The namespaces `SimplePie\Tests\Integration` and `SimplePie\Tests\Unit` contains only tests so far. Can we move this helper class into the `SimplePie\Tests\Fixtures` namespace?

jtojnar: Makes sense, though I hope to drop it tomorrow when, hopefully [the fix](https://github.com/donatj/mock-webserver/pull/75) will be merged upstream.

Art4 (approved):

PR #926, opened 2025-07-27
Fix parse_date() return type

Alkarex:
Corrected to `@return int|false`


jtojnar:
Thanks. Double checked that it matches all the `Parse\Date::parse*` methods.

Issue #190, opened 2012-05-14
Create new SimplePie_Feed class


rmccue:
At the moment, `SimplePie_Core` is both the core feature handler as well as the feed level handler. This has a whole bunch of problems (multifeeds, e.g.) which can be solved by splitting them. See [SPEP #100](https://github.com/rmccue/SPEP/blob/master/docs/spep-0100.markdown) for more information.

`SimplePie_Core::get_items()` would give the merged items as normal, however feed-level data should give a deprecated warning (but still return a value for backwards compatibility).

A new method should also be added to get the array of feeds for looping individually.

PR #930, opened 2025-07-28
expose headers on failure

jtojnar:


src/SimplePie.php
```diff
@@ -2115,6 +2118,27 @@ public function status_code()
         return $this->status_code;
     }
 
+    /**
+     * Get HTTP headers of the last request
+     *
+     * @return array<string, string[]>
+     */
+    public function get_http_headers(): array
```
jtojnar, line 2126: Not sure if we want to have this. Maybe exposing `Response` will be better.

Art4: I think this new method is a reasonable compromise. Once we have exposed the `Response` class or implemented #190, we will be able to deprecate many methods.

tests/Integration/SimplePieTest.php
```diff
@@ -147,6 +180,7 @@ public function testSimplePieReturnsCorrectStatusCodeFromServerResponse(): void
 
         $this->assertFalse($return);
         $this->assertSame(429, $simplepie->status_code());
+        $this->assertSame(['3600'], $simplepie->get_http_header('Retry-After'));
```
Art4, line 183: Maybe adding a test for `SimplePie::get_http_headers()` as well?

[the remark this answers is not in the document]
jtojnar: Actually, looking at the headers, most of them are set by PHP test server itself so we probably cannot rely on them remaining consistent:

```php
'Host' => Array &1 [
   0 => '127.0.0.1:39477',
],
'Date' => Array &2 [
   0 => 'Fri, 01 Aug 2025 08:57:44 GMT',
],
'Connection' => Array &3 [
   0 => 'close',
],
'X-Powered-By' => Array &4 [
   0 => 'PHP/8.4.8',
],
'Retry-After' => Array &5 [
    0 => '3600',
],
'Content-type' => Array &6 [
   0 => 'text/html; charset=UTF-8',
],
```

PR #932, opened 2025-08-01
Bump GitHub Actions

Alkarex:
Update GitHub Actions as well as their corresponding PHP version.
Note that the *Test single-file build* fails with PHP 8.4, so kept on PHP 8.3 for now (should likely be in another PR anyway).


jtojnar:
Thanks.

PR #933, opened 2025-08-01
Use correct language as identifier in Heredoc strings

Alkarex:
To allow correct syntax highlighting in IDEs, for e.g. XML, HTML, JS instead of showing the code snippets as plaintext


Alkarex:
This has annoyed me a bit for a while, so here it is fixed.
Comparison before (left) / after (right):

image

src/Misc.php
```diff
@@ -2114,7 +2114,7 @@ function embed_flv(width, height, link, placeholder, loop, player) {
 function embed_wmedia(width, height, link) {
 	document.writeln('<embed type="application/x-mplayer2" src="'+link+'" autosize="1" width="'+width+'" height="'+height+'" showcontrols="1" showstatusbar="0" showdisplay="0" autostart="0"></embed>');
 }
-END;
+JS;
```
Alkarex, line 2117: Example of JavaScript

tests/SanitizeTest.php
```diff
@@ -15,14 +15,14 @@ public function testSanitize(): void
...
 &lt;head&gt; &amp; &lt;body&gt; /\ ' === ' &amp; " === ". Sbohem bez šátečku! Тут был Лёха.
-EOT
+HTML
             ,
             $sanitize->sanitize(
-                <<<EOT
+                <<<HTML
 &#60;head&#62; &amp; &lt;body&gt; /\ ' === &apos; &#38; " === &quot;. Sbohem bez šátečku! Тут был Лёха.<script>alert('XSS')</script>
-EOT
+HTML
```
Alkarex, line 25: Example of HTML

PR #939, opened 2025-09-14
Upgrade to PHPStan 2

Alkarex:
https://github.com/phpstan/phpstan/blob/2.0.x/UPGRADING.md
PHPStan has improved a lot since version 1.12 used by SimplePie until now.
Take advantage of explicit PHPStan phpVersion range: define explicit range of minimum and maximum PHP version supported. Useful e.g. for function signatures, which have changed over time https://phpstan.org/config-reference#phpversion
Do not run PHPStan for PHP < 8.0 but the tests are covered by phpVersion min, so it would actually be enough to run PHPStan once on any supported PHP version (keep that optimisation for another PR).
Cannot support PHPUnit >= 10 at the same time as PHP 7.2 for now.


tests/Unit/SimplePieTest.php
```diff
@@ -3145,6 +3145,6 @@ public function test_get_title(string $data, string $expected): void
         $feed->enable_cache(false);
         $feed->init();
 
-        $this->assertSame($expected, $feed->get_title());
+        self::assertSame($expected, $feed->get_title());
```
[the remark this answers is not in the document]
Alkarex: Sure, done in https://github.com/simplepie/simplepie/pull/940
If that one gets merged first, then the current PR should automatically be updated accordingly. If not, I can do it manually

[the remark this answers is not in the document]
Alkarex: P.S. I was using PHPStan in strict mode (which reports those errors) when preparing a part of this PR, but strict mode is not a part of this PR.

composer.json
```diff
@@ -69,7 +70,7 @@
         "coverage": "phpunit --coverage-html=.phpunit.cache/code-coverage",
         "cs": "php-cs-fixer fix --verbose --dry-run --diff",
         "fix": "php-cs-fixer fix --verbose --diff",
-        "phpstan": "phpstan analyze --memory-limit 512M",
```
jtojnar, line 72: Without this, I get the following when running `composer phpstan` locally:

```
     Child process error: PHPStan process crashed because it reached configured PHP memory limit: 128M  
     Increase your memory limit in php.ini or run PHPStan with --memory-limit CLI option.               
      while running parallel worker                                                                     
```

Alkarex: Was it with PHPStan 2?
We can just add this parameter again, but my experience has been that memory requirements have decreased with this new version, and I have not experienced a need for it in SimplePie or FreshRSS.
But some OS seem to be less efficient though. Are you maybe running on something else than Linux?

jtojnar: Yes, on this branch after running `composer install`. I am on Linux. Maybe your environment has higher than the default memory limit? https://github.com/php/php-src/blob/e6c07b004b654dcc2362bd0ea29c269e4a531a5a/php.ini-production#L430

phpstan.dist.neon
```diff
@@ -1,65 +1,25 @@
...
-
-        -
-            message: '(^Parameter \$parser of method SimplePie\\Parser::(tag_open|cdata|tag_close)\(\) has invalid type XMLParser\.$)'
-
-            count: 3
-            path: src/Parser.php
-            # Only occurs on PHP ≤ 7.4
-            reportUnmatched: false
+    excludePaths:
+        analyseAndScan:
```
jtojnar, line 15: Why is this needed? None of these are listed under `paths`.

.github/workflows/ci.yml
```diff
@@ -82,7 +86,9 @@ jobs:
       - name: Run unit tests
         run: composer test
 
+      # It would be sufficient to run PHPStan with one supported PHP version
```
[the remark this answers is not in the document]
Alkarex: From PHPStan 2.0, which supports a PHP version range, variations between PHP versions on the whole range are accounted for, for instance PHP functions that have changed signature over time. I am not aware cases where it would make a difference to run PHPStan on multiple PHP versions, since the same rules will be checked (as opposed to PHPStan 1.x, where it could make a difference)

composer.json
```diff
@@ -33,8 +33,9 @@
         "donatj/mock-webserver": "^2.7",
         "friendsofphp/php-cs-fixer": "^2.19 || ^3.8",
         "mf2/mf2": "^0.5.0",
-        "phpstan/phpstan": "~1.12.2",
-        "phpunit/phpunit": "^8 || ^9 || ^10",
+        "phpstan/phpstan": "~2.1.25",
+        "phpstan/phpstan-phpunit": "~2.0.7",
+        "phpunit/phpunit": "^8 || ^9",
```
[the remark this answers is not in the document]
Alkarex: Additional debugging welcome (but maybe not urgent). I am facing a symbol discovery issue when trying to support both PHPUnit >= 10 and PHP 7.2 at the same time

phpstan.dist.neon
Alkarex: Right, mainly because it may be interesting to run PHPStan manually on selected parts of the code base.
Furthermore (more tests welcome), I have the impression that `paths` is not sufficient and files outside those paths are also scanned based on other discovery rules

demo/index.php
```diff
@@ -122,7 +122,13 @@
                 echo '<div class="sp_errors">' . "\r\n";
 
                 // ... and display it.
-                echo '<p>' . htmlspecialchars($feed->error()) . "</p>\r\n";
+                if (is_string($feed->error())) {
```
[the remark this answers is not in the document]
jtojnar: Deprecating it in #948 along with some other demo cleanups. Including your that make sense outside of PHPStan bump.

phpstan.dist.neon
jtojnar: Using `echo '<?php \PHPStan\dumpType(1 + 1);' | tee .git/foo.php build/foo.php vendor/foo.php` and `composer phpstan -- -v .`, I verified that `.git` is not necessary while the other two are.

utils/PHPStan/RegistryCallMethodReturnTypeExtension.php
jtojnar, line 1: Handling that properly in https://github.com/simplepie/simplepie/pull/969

PR #940, opened 2025-09-14
PHPUnit assert functions are static

Alkarex:
Flagged by PHPStan (strict mode). Found during https://github.com/simplepie/simplepie/pull/939

I am actually surprised that the [PHPUnit documentation](https://docs.phpunit.de/en/12.3/assertions.html) proposes to use either `$this->assertTrue()` or `self::assertTrue()`. A static method should be called statically (I could not even find in the official PHP documentation anything that suggests that it should be allowed otherwise, but it does still work so far):
* https://phpstan.org/error-identifiers/staticMethod.dynamicCall
* https://github.com/phpstan/phpstan/issues/514

### DIFF
--- a/tests/BCTest.php
+++ b/tests/BCTest.php
@@ -17,7 +17,7 @@ class BCTest extends TestCase
      */
     public function test_class_SimplePie_Core_exists(): void
     {
-        $this->assertTrue(class_exists('SimplePie_Core'));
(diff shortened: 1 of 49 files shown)


jtojnar (approved):
Makes sense to me.

Art4 (approved):

PR #941, opened 2025-09-15
Fix typos in @throws class names in Cache/Psr16.php

samwilson:
The correct way to separate multiple class names is with `|` and not `&`.

This updates `@throws InvalidArgumentException&Throwable` to `@throws InvalidArgumentException|Throwable`.

### DIFF
--- a/src/Cache/Psr16.php
+++ b/src/Cache/Psr16.php
@@ -48,7 +48,7 @@ final class Psr16 implements DataCache
      *
      * @return array|mixed The value of the item from the cache, or $default in case of cache miss.
      *
-     * @throws InvalidArgumentException&Throwable
+     * @throws InvalidArgumentException|Throwable
      *   MUST be thrown if the $key string is not a legal value.
      */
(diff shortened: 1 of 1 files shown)


Art4:
`|` describes [Union Types](https://php.watch/versions/8.0/union-types), `&` describes [Intersection Types](https://php.watch/versions/8.1/intersection-types).

Using Intersection Type declaration we are making clear that the throwing exception will implement both `Psr\SimpleCache\InvalidArgumentException` and `Throwable`.

Art4:
The `Psr\SimpleCache\InvalidArgumentException` extends `Psr\SimpleCache\CacheException` from PSR-16 simple cache. 

The `CacheExeption` [implements `\Throwable` interface only since PSR-16 v2.0.0](https://www.php-fig.org/psr/psr-16/meta/#81-throwable) but SimplePie supports PSR-16 v1, v2 and v3. Once SimplePie drops support for PHP 7.4, we can require PSR-16 v2 and can than remove the `&Throwable` intersection.

PR #949, opened 2025-09-29
PHP 8.5: Fix "Using null as an array offset" deprecations

swissspidy:
This is a new deprecation in PHP 8.5. See https://wiki.php.net/rfc/deprecations_php_8_5 / https://github.com/php/php-src/commit/49e3956b703b4aeb24c7a42ae3ee89e2a339560d


jtojnar (commented):
Thanks for the fix. Confirmed that this is indeed a problem with php-src master:

```php
<?php
$foo = [];
$scheme = null;
var_dump(isset($foo[$scheme]));
// Deprecated: Using null as an array offset is deprecated, use an empty string instead in /in/Udd6D on line 6
// bool(false)
```

jtojnar:
Thank you.

PR #951, opened 2025-10-01
Apply HTTPS policy consistently

Alkarex:
Context: We observed some issues with some duplicate entries from feeds of WordPress.org due to inconsistencies in their HTTP vs. HTTPS. The culprit is:
* https://github.com/Automattic/pushpress/issues/15

SimplePie has a mechanism for enforcing HTTPS on selected domains (none by default), but this mechanism was not applied everywhere, lacking in particular item ID.

Downstream:
* https://github.com/FreshRSS/FreshRSS/pull/8037
* https://github.com/FreshRSS/FreshRSS/issues/5151


src/Item.php
```diff
@@ -2353,7 +2365,7 @@ public function set_sanitize(Sanitize $sanitize): void
         $this->sanitize = $sanitize;
     }
 
-    protected function get_sanitize(): Sanitize
+    public function get_sanitize(): Sanitize
```
jtojnar: I would prefer `Source` to have its own instance of `Sanitize`. The more public API, the less we can change in the future.

src/Item.php
```diff
@@ -201,18 +201,23 @@ public function get_feed()
...
             } elseif ($return = $this->get_item_tags(\SimplePie\SimplePie::NAMESPACE_DC_10, 'identifier')) {
-                return $this->sanitize($return[0]['data'], \SimplePie\SimplePie::CONSTRUCT_TEXT);
+                $guid = $this->sanitize($return[0]['data'], \SimplePie\SimplePie::CONSTRUCT_TEXT);
             } elseif (isset($this->data['attribs'][\SimplePie\SimplePie::NAMESPACE_RDF]['about'])) {
-                return $this->sanitize($this->data['attribs'][\SimplePie\SimplePie::NAMESPACE_RDF]['about'], \SimplePie\SimplePie::CONSTRUCT_TEXT);
+                $guid = $this->sanitize($this->data['attribs'][\SimplePie\SimplePie::NAMESPACE_RDF]['about'], \SimplePie\SimplePie::CONSTRUCT_TEXT);
+            }
+            if ($guid !== '') {
+                // If the ID looks like a URL, apply HTTPS policy to it.
+                return $this->get_sanitize()->https_url($guid);
```
jtojnar: This will probably be a BC break.

Also not sure it makes sense – guid should be stable.

Alkarex: I can remove the GUID part from the PR if you prefer.
In the case of WordPress, GUID is not stable regarding HTTP vs. HTTPS, which leads to duplicates.
I have not observed any feed where it would be a problem to normalise the HTTP(S) of their GUIDs.

jtojnar: Yeah, that would be preferable.

IMO, Wordpress is wrong here, it should keep the old guid across protocol, domain and URI structure changes. If it is impractical to guarantee that, they could have always used UUID or something.

The problem would occur in feeds that still contain `http://` items – suddenly, they would be duplicated in the feed reader.

**Edit:** Out of 372 feeds I am subscribed to, three feeds contain both HTTP and HTTPS links as part of the guid:

- https://blog.mecheye.net/feed/ (note this is a Wordpress blog so it appears wordpress does support immutable guids)
- https://planet.gnome.org/rss20.xml
- https://trofi.github.io/feed/rss.xml

The others either use protocol-independent GUIDs like http://0pointer.net/blog/index.atom, switched all their guids to https (causing duplicates) or only keep recent items in the feed.

src/Item.php
Alkarex: Done https://github.com/simplepie/simplepie/pull/951/commits/1b73339683e3eb696bf1bb72474337c442668ccb

src/Item.php
Alkarex: Done https://github.com/simplepie/simplepie/pull/951/commits/44e099d52074fe980f36787a4ce9d7009193e876

PR #955, opened 2025-11-06
Exclude some more unwanted files from the package

johnbillion:
These files aren't needed in the Composer package.


jtojnar:
Thanks, sounds good:

- While the PHPStan extension might be useful to dependents, we consider it unstable so not exposing it is probably fine. There seem to be no third-party users on GitHub at least.
- The `db.sql` contains DDL statements for multiple databases so no-one should be using it in automated way.
- Dependents should not need to perform static analysis.

PR #957, opened 2025-11-25
Fixes before phpstan2

Alkarex:
Since https://github.com/simplepie/simplepie/pull/939 has been waiting for a long time, here is a PR still using PHPStan 1 (so it should be easier to merge), but fixing many problems caught by PHPStan 2.

In other words, this PR is a subset of https://github.com/simplepie/simplepie/pull/939


jtojnar:
Thanks.

Issue #961, opened 2026-01-01
Issue with websites sending invalid Content-Encoding headers


mmeier86:
This issue is related to FreshRSS/FreshRSS#8374

I came across an issue with blogs hosted on alternative S3 implementations (Ceph, in this case) and being uploaded with Hugo's `deploy` command. The effect of this setup is that files from that bucket get the `Content-Encoding: aws-chunked` header, which is then also send out when the file, in this case the `index.xml` containing the RSS feed, is requested by a client.

This leads to problems in FreshRSS, using SimplePie, as cURL only supports certain specific values for the `Content-Encoding` header. SimplePie currently already has a workaround for this issue:
https://github.com/simplepie/simplepie/blob/92bc715d126ae53d98577cdb5a6903af47b9138e/src/File.php#L141-L144

The `none` value set in the current code does not seem to be officially supported by cURL, which only accepts `""`, which sends an `Accept-Encoding` header with all supported encodings, or a specific encoding, namely `identity`, `deflate`, `gzip`, `br` and `zstd`. With `none` set, the server might still send the `Content-Encoding: aws-chunked` invalid header, in which case curl errors out, as it doesn't know what to do with this encoding.

My proposal is to change the value from `'none'` to `null`. This disables any attempt at decompressing the response body on cURL's side. At least in my tests in FreshRSS, this allowed SimplePie to successfully fetch the `index.xml` file and extract the RSS feed from it, even when the server sends the invalid `Content-Encoding` header.

I will provide a PR which makes the proposed change from `'none'` to `null` in a moment.

PR #962, opened 2026-01-01
Disable cURL encoding/compression handling when a bad Content-Encoding header is received

mmeier86:
This PR relates to #961 and FreshRSS/FreshRSS#8374.

When cURL has already returned a bad encoding error, this is a good indication that the server ignores `Accept-Encoding` headers, as the `CURLOPT_ENCODING` option is already set to `''` for the first request. This results in cURL sending an `Accept-Encoding` header with all supported encodings (compression algorithms, in practice) already in the first request. And the fact that this request results in a `CURLE_BAD_CONTENT_ENCODING` error indicates that the server ignored that header already. So trying again, but this time with `CURLOPT_ENCODING = 'none'` seems unlikely to result in success?

Instead, my proposal here is to set `CURLOPT_ENCODING` to `null`, which tells cURL to disable compression handling altogether and to ignore the bad `Content-Encoding` header send by the server.

I was able to confirm that this works in tests with FreshRSS against a test setup where the server sends a valid `index.xml` with an RSS feed, but with the `Content-Encoding` header set to `aws-chunked`. With `CURLOPT_ENCODING = 'none'`, the fetch fails and the body is empty, leading to FreshRSS not being able to add the feed. But with `CURLOPT_ENCODING = null` set, the fetch succeeds and the feed can be added.

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -110,36 +110,7 @@ class File implements Response
             }
             if (!$force_fsockopen && function_exists('curl_exec')) {
                 $this->method = \SimplePie\SimplePie::FILE_SOURCE_REMOTE | \SimplePie\SimplePie::FILE_SOURCE_CURL;
-                $fp = curl_init();
-                $headers2 = [];
-                foreach ($headers as $key => $value) {
-                    $headers2[] = "$key: $value";
-                }
-                if (isset($curl_options[CURLOPT_HTTPHEADER])) {
-                    if (is_array($curl_options[CURLOPT_HTTPHEADER])) {
-                        $headers2 = array_merge($headers2, $curl_options[CURLOPT_HTTPHEADER]);
-                    }
-                    unset($curl_options[CURLOPT_HTTPHEADER]);
-                }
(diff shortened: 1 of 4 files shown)


mmeier86:
I'm not exactly fluent in PHP, but I will give it a good try.

jtojnar:
Looks like setting NULL does not work on PHP < 8.1 because of https://github.com/php/php-src/issues/11433, which leads to curl returning:

    string(113) "cURL error 61: Unrecognized content encoding type. libcurl understands deflate, gzip, br, zstd content encodings."

jtojnar:
I managed to work around it by creating a new curl handle from scratch but PHPStan does not like that on older versions of PHP where `curl_init` returns `resource` instead of `CurlHandle`:

https://github.com/simplepie/simplepie/compare/master...jtojnar:simplepie:disable-curl-encoding-handling-for-bad-content-encoding?expand=1

jtojnar (approved):
Thanks.

Art4 (approved):
@mmeier86 Thanks for your contribution!

@jtojnar nice solution for curl_init() and PHPStan, thanks!

PR #964, opened 2026-01-14
fix: Return null instead of error or today date for invalid date formats

arraintxo:
Fixes date parsing behavior when encountering invalid or localized date formats.

**Before (PHP 7.4):**
- Invalid dates could throw errors/warnings due to `strict_types`

**Before (PHP 8.*):**
- Invalid dates returned current date when not properly parsed (`date` uses current time when passed null)

**After:**
- Invalid dates gracefully return null
- Consistent behavior for different PHP versions

**Changes:**
- Improved error handling in date parsing
- Added test coverage for RFC 2822 formatted dates
- Added test for custom format parameter

### DIFF
--- a/tests/Unit/ItemTest.php
+++ b/tests/Unit/ItemTest.php
@@ -1318,7 +1318,7 @@ XML
                 ,
                 1168531200,
             ],
-            'Test RFC 2822/ RFC 5322 formatted date' => [
+            'Test RFC 2822 formatted date' => [
 <<<XML
 <rss version="2.0">
 	<channel>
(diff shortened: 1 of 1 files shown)


src/Item.php
```diff
@@ -672,7 +672,7 @@ public function get_date(string $date_format = 'j F Y, g:i a')
                     return $this->data['date']['parsed'];
 
                 default:
-                    return date($date_format, $this->data['date']['parsed']);
+                    return $this->data['date']['parsed'] ? date($date_format, $this->data['date']['parsed']) : null;
```
jtojnar: Could you use explicit null check so that `1970-01-01 00:00:00` does not return `null`?

```suggestion
                    return $this->data['date']['parsed'] !== null ? date($date_format, $this->data['date']['parsed']) : null;
```

jtojnar (commented):
Thanks, this looks great, just few notes.

PR #974, opened 2026-03-29
New function get_icon_url() for feed favicon

Alkarex:
Only atom:icon and a square rss2:image can be used to represent a favicon.

The existing get_image_url() is more for logos / banners, not favicon.

Examples of real-world feeds:
* https://feedpress.me/frandroid
* https://www.lesnumeriques.com/rss.xml

Downstream:
* https://github.com/FreshRSS/FreshRSS/issues/5518
* https://github.com/FreshRSS/FreshRSS/pull/8633


jtojnar (approved):
Thanks, the proposed API looks reasonable to me.

The code is good and well tested.

Art4 (approved):
👍

PR #975, opened 2026-04-17
Fix int types for enclosures

Alkarex:
channels, duration were wrongly parsed as string instead of int.

```
PHP Fatal error:  Uncaught TypeError: SimplePie\Enclosure::__construct():
Argument #12 ($duration) must be of type ?int, string given in simplepie/src/Enclosure.php:199
```

And add tests (there were none for those attributes)

Downstream issue:
* https://github.com/FreshRSS/FreshRSS/issues/8701


jtojnar (approved):
Thanks. Ideally, we would have `CONSTRUCT_POSITIVE_INT` or even `CONSTRUCT_DURATION` that would return null or error instead of 0 but this is a nice and necessary improvement.

PR #978, opened 2026-05-25
Fix wrong player_parent logic leading to invalid type

Alkarex:
```
PHP Fatal error: Uncaught TypeError:
SimplePie\Enclosure::__construct(): Argument #20 ($player) must be of type ?string, array given
in /opt/freshrss/lib/simplepie/simplepie/src/Enclosure.php:199
```

This happens when there is no`url` parameter.

Example of feed:
* https://feeds.feedburner.com/crunchyroll/rss/anime?lang=deDE

Downstream issue:
* https://github.com/FreshRSS/FreshRSS/issues/8892

### DIFF
--- a/src/Item.php
+++ b/src/Item.php
@@ -1224,13 +1224,14 @@ class Item implements RegistryAware
             }
 
             // PLAYER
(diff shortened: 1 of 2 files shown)


jtojnar:
Thanks. Looks great.

PR #979, opened 2026-06-10
PHP 8.5: Fix null warning in IRI

Alkarex:
In PHP 8.5, `empty($my_array[null])` generates a warning:
> PHP Deprecated:  Using null as an array offset is deprecated, use an empty string instead

### DIFF
--- a/src/IRI.php
+++ b/src/IRI.php
@@ -1023,7 +1023,7 @@ class IRI
         }
         if ($this->ipath !== '') {
             $iri .= $this->ipath;
-        } elseif (!empty($this->normalization[$this->scheme]['ipath']) && $iauthority !== null && $iauthority !== '') {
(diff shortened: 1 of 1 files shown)


jtojnar (approved):
Thanks.

This can be reproduced using e.g. `new SimplePie\IRI("//foo")->get_iri()`.

I wondered if this is a BC break to a subclass that relies on null being cast to an empty string for normalisation of relative URIs. But that can probably be disregarded, as we normalise the scheme field to null everywhere and pre-empt the cast in the other instances using null checks:

- https://github.com/simplepie/simplepie/blob/4cc2492a087e4361f647647c095a44389aa7c413/src/IRI.php#L173
- https://github.com/simplepie/simplepie/blob/4cc2492a087e4361f647647c095a44389aa7c413/src/IRI.php#L626

PR #981, opened 2026-06-23
Fix typo "contrbutor" in get_contributor() docblocks

### LINKED ISSUE #980: Typo "contrbutor" in get_contributor() docblocks (Item.php, SimplePie.php)
## Description

There is a small typo — `contrbutor` instead of `contributor` — in the `@param` docblock of the `get_contributor()` method, in two files:

- [`src/Item.php` line 464](https://github.com/simplepie/simplepie/blob/master/src/Item.php#L464)
- [`src/SimplePie.php` line 2703](https://github.com/simplepie/simplepie/blob/master/src/SimplePie.php#L2703)

```php
* @param int $key The contrbutor that you want to return. Remember that arrays begin with 0, not 1
```

Should read:

```php
* @param int $key The contributor that you want to return. Remember that arrays begin with 0, not 1
```

## Context

This was originally noticed downstream in WordPress core, which bundles the SimplePie library. As suggested by a WordPress core committer, the fix belongs upstream here in the SimplePie repository: https://core.trac.wordpress.org/ticket/65499

I'll open a PR fixing both occurrences.

itzmekhokan:
## Description

Fixes a typo — `contrbutor` → `contributor` — in the `@param` description of the `get_contributor()` method in two files:

- `src/Item.php`
- `src/SimplePie.php`

Documentation-only change; no functional impact.

Fixes #980

## Context

Originally spotted downstream in WordPress core (which bundles SimplePie). Per a WordPress core committer's suggestion, the fix is being made upstream here: https://core.trac.wordpress.org/ticket/65499

### DIFF
--- a/src/Item.php
+++ b/src/Item.php
@@ -461,7 +461,7 @@ class Item implements RegistryAware
      * Get a contributor for the item
      *
      * @since 1.1
-     * @param int $key The contrbutor that you want to return.  Remember that arrays begin with 0, not 1
+     * @param int $key The contributor that you want to return.  Remember that arrays begin with 0, not 1
      * @return \SimplePie\Author|null
      */
     public function get_contributor(int $key = 0)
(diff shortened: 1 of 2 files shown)


jtojnar (approved):
Thanks

PR #982, opened 2026-06-25
Fix cURL test on Alpine Linux

Alkarex:
Problem observed on Alpine 3.24 with PHP 8.5.7 and libcurl 8.20.0:
```
1) SimplePie\Tests\Integration\SimplePieTest::testSimplePieReturnsCorrectStatusCodeOnServerConnectionError
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'cURL error 6: Could not resolve host: example.invalid'
+'cURL error 6: Could not resolve host: example.invalid (Domain name not found)'
```


tests/Integration/SimplePieTest.php
```diff
@@ -163,7 +163,8 @@ public function testSimplePieReturnsCorrectStatusCodeOnServerConnectionError():
 
         self::assertFalse($return);
         self::assertSame(0, $simplepie->status_code());
-        self::assertSame('cURL error 6: Could not resolve host: example.invalid', $simplepie->error());
+        self::assertIsString($simplepie->error());
```
jtojnar: Maybe add a comment `// For PHPStan` like in the other fixes for this.

jtojnar (approved):
Thanks.

tests/Integration/SimplePieTest.php
Alkarex: Done

PR #984, opened 2026-06-25
CI/CD: Update to PHP 8.5 for test-compiled

Alkarex:
PHP 8.3 is no longer in active support.
Edit: problem with MF2, which I had not anticipated when starting the PR:
* https://github.com/microformats/php-mf2/pull/269


.github/workflows/ci.yml
```diff
@@ -92,7 +92,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ['8.3']
+        php: ['8.4']
```
jtojnar: Should we perhaps run it on the latest version? It looks like you kept it back because of some bug in 4c53bd01add1ecefa5c593c84acf6d2b1fe56de5 but that appears to no longer be relevant.

jtojnar (approved):

.github/workflows/ci.yml
Alkarex: As you prefer

Alkarex: Done

